PR #3, opened 2016-06-30
ENH: Adding in balance basis with tests

mortonjt:
Creates an orthonormal basis given a binary bifurcating tree.


gneiss/balances.py
```diff
@@ -0,0 +1,115 @@
...
+import numpy as np
+from skbio.stats.composition import clr_inv
+from collections import OrderedDict
+
+
+def _balance_basis(tree_node):
+    """ Helper method for calculating balance basis
+    """
+
+    n_tips = sum([n.is_tip() for n in tree_node.traverse()])
```
antgonza: do you think this line can be incorporated in the _count_matrix logic? Just wondering if we can save some loops.

gneiss/balances.py
```diff
@@ -0,0 +1,115 @@
...
+    counts = _count_matrix(tree_node)
+    counts = OrderedDict([(x, counts[x])
```
antgonza: Now, wondering if is worth having this logic within _count_matrix, what do you think?

gneiss/balances.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import division
+import numpy as np
+from skbio.stats.composition import clr_inv
+from collections import OrderedDict
+
+
+def _balance_basis(tree_node):
```
antgonza: Yes, it seems like all these lines can be within _count_matrix but perhaps I'm missing something obvious. Could you clarify?

gneiss/balances.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def balance_basis(tree_node):
+    """
+    Determines the basis based on tree
+
+    Parameters
+    ----------
+    treenode : skbio.TreeNode
+        Phylogenetic tree.  Must be a strictly bifurcating tree
```
antgonza: Can you add a test making sure that the tree is bifurcating?

gneiss/balances.py
mortonjt: So have `_count_matrix` return `n_tips` to avoid extra looping?

I like it!

gneiss/balances.py
mortonjt: 👍

gneiss/balances.py
```diff
@@ -0,0 +1,115 @@
...
+    basis = clr_inv(basis)
+    return basis, nodes
+
+
+def _count_matrix(treenode):
+    nodes = list(treenode.levelorder(include_self=True))
+    # fill in the dictionary
+    counts = OrderedDict()
+    columns = ['k', 'r', 'l', 't', 'tips']
+    for n in nodes:
```
antgonza: What about using defaultdicts to avoid this init step?

gneiss/balances.py
mortonjt: Although, there will be an extra loop required anyways, since `is_tip()` will need to be evaluated over all of the nodes anyways.  So there won't be any savings in this case.  Does that make sense?

gneiss/balances.py
```diff
@@ -0,0 +1,115 @@
...
+        if n not in counts:
+            counts[n] = {}
+        for c in columns:
+            counts[n][c] = 0
+
+    # fill in r and l
+    for n in nodes[::-1]:
```
antgonza: Can you add a comment of why ::-1?

gneiss/balances.py
```diff
@@ -0,0 +1,115 @@
...
+            lchild = n.children[0]
+            rchild = n.children[1]
+            counts[n]['r'] = counts[rchild]['tips']
+            counts[n]['l'] = counts[lchild]['tips']
+            counts[n]['tips'] = counts[n]['r'] + counts[n]['l']
+        else:
+            raise ValueError("Not a strictly bifurcating tree!")
+
+    # fill in k and t
+    for n in nodes:
```
antgonza: This will also benefit from using defaultdicts.

gneiss/balances.py
antgonza: I'm guessing you are already doing that when you evaluate 't', right?

antgonza:
Some comments, will review again when they are addressed and test pass.

gneiss/balances.py
mortonjt: You're suggesting to move this all code block into `_count_matrix` correct?

We could do that.  The main reason why I wanted to keep them separate was because I wanted to be able to test these modules separately.  What do you think about that?

gneiss/balances.py
mortonjt: 👍

gneiss/balances.py
mortonjt: Because it is filling it in backwards level order.  Will add comment.

gneiss/balances.py
antgonza: That makes sense if you think is important to test the midpoint values (for example, if they change the final output will not change) and you believe you can reuse that function. So I leave it up to you.

gneiss/balances.py
mortonjt: Here are the actual timings of `_balance_basis` as is

```python
10000 loops, best of 3: 43.5 µs per loop
```
And here are the timings if I put in the `n_tips` calculation within the second loop of the `_count_matrix` function.
```python
10000 loops, best of 3: 37.7 µs per loop
```

So not much difference at all.  I don't have a strong preference, so we can just do it the way you suggested.

gneiss/balances.py
mortonjt: Ok, I actually take that back.

The reason why I'm doing the OrderedDict calculation outside of `_count_matrix` was to ease the process of testing `_count_matrix`.  I'm reverting back the change - hope that is ok.

gneiss/balances.py
mortonjt: Does [this](https://github.com/biocore/gneiss/pull/3/files#diff-1079b10233352c49b5ec47ae9b0644e5R46) address your concerns?

gneiss/balances.py
mortonjt: Ok, I also take that back.

The reason why we want to have an `OrderedDict` is because the ordering is extremely important - we need to ensure level order.  If we use a straight-up defaultdict, we wouldn't necessarily have this ordering.  Does this make sense?

gneiss/balances.py
mortonjt: Right.  But would defaultdicts maintain the level-ordering?  If not, then this would mess things up.

gneiss/balances.py
antgonza: OK, just note that in this code only `counts` is an OrderedDict but the contents are not:
```python
>>> counts = OrderedDict()
>>> type(counts)
<class 'collections.OrderedDict'>
>>> counts['label'] = {}
>>> counts['label']['A'] = 1
>>> type(counts)
<class 'collections.OrderedDict'>
>>> type(counts['label'])
<type 'dict'>
```

gneiss/balances.py
wasade: what size trees are you looking to scale too out of curiosity, and will you be performing a large number of operations on these trees?

gneiss/balances.py
josenavas: Not sure if it is worth, but though I will point this out: http://stackoverflow.com/a/6190500/3746629

mortonjt: Thanks for linking!  If this were a data structure that will be continuously used, I'd think about it.

I don't think it is necessary in this context, since this `OrderedDict` is only used to store intermediate results.  The only thing that matters is the matrix, and the nodes that are returned.

gneiss/balances.py
mortonjt: As large as we can visualize them.  From prior experience, no more than 4000 tips - it starts to become very difficult to visualize anything past that.

gneiss/balances.py
mortonjt: Ok, cool.  I'll leave as is for the sake of testing.

gneiss/balances.py
wasade: kk

mortonjt:
Tests pass!  All comments so far have been addressed.

gneiss/balances.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    basis, nodes = _balance_basis(tree_node)
+    basis = clr_inv(basis)
+    return basis, nodes
+
+
+def _count_matrix(treenode):
+    n_tips = 0
+    nodes = list(treenode.levelorder(include_self=True))
+    # fill in the Ordered dictionary.  Note that the
```
antgonza: Extra space between `.` and Note.

antgonza:
One minor comment and could you also add flake8 to travis? Thanks!

gneiss/balances.py
mortonjt: Done

mortonjt:
Space removed.  Flake8 and pep8 are enabled within `Makefile`.
Want to merge this?

antgonza:
Thanks @mortonjt

PR #4, opened 2016-07-04
ENH: Adding in balance plot

mortonjt:
There is the skeleton for the balance plot - this is just used for plotting values on the internal nodes. 

There are few things that I would like some input for 

1. Testing - the only way that I know how to test this code is using IPython notebooks.  This is because the actual ETE visualizations aren't tested.  But IPython notebooks aren't ideal to review / modify, due to the underlying json.  Another option is to use [ipymd](https://media.readthedocs.org/pdf/ipymd/latest/ipymd.pdf) to render as markdown.  Or we can use Sphinx to display plots.  I'm not sure what the best way forward on this is
2. The balances parameter in `balanceplot` right now only accepts a `np.array` as input, and assumes level ordering.  I think this is good enough for the first pass.  Eventually I'd like to pass in a `pd.Series` to make this easier to use.  The problem with this is the internal nodes within the tree aren't labeled most of the time.
3. Display layouts.  All of the ETE display parameters are enforced within a `layout` function.  I think the default layouts that I have atm are good enough for the first pass, but I'd like to hear some ideas on how to improve the usability of this.

### DIFF
--- a/gneiss/balances.py
+++ b/gneiss/balances.py
@@ -2,6 +2,7 @@ from __future__ import division
 import numpy as np
 from skbio.stats.composition import clr_inv
 from collections import OrderedDict
+from ete3 import Tree, TreeStyle, faces, AttrFace, CircleFace, BarChartFace
 
 
 def _balance_basis(tree_node):
--- a/gneiss/balances.py
+++ b/gneiss/balances.py
@@ -129,3 +129,107 @@ def _count_matrix(treenode):
             counts[n]['k'] = counts[n.parent]['k'] + counts[n.parent]['r']
             counts[n]['t'] = counts[n.parent]['t']
     return counts, n_tips
+
+
+def default_layout(node):
+    if node.is_leaf():
+        # Add node name to leaf nodes
+        N = AttrFace("name", fsize=14, fgcolor="black")
+
+        faces.add_face_to_node(N, node, 0)
+    if "weight" in node.features:
+        # Creates a sphere face whose size is proportional to node's
+        # feature "weight"
+        C = CircleFace(radius=node.weight, color="Red", style="sphere")
+        # Let's make the sphere transparent
+        C.opacity = 0.5
+        # Rotate the faces by 90*
+        C.rotation = 90
+        # And place as a float face over the tree
+        faces.add_face_to_node(C, node, 0, position="float")
+
+
+def barchart_layout(node, name='name',
+                    width=20, height=40,
+                    colors=None, min_value=0, max_value=1,
+                    fsize=14, fgcolor="black"):
+    if colors is None:
+        colors = ['#0000FF']
+    if node.is_leaf():
+        # Add node name to leaf nodes
+        N = AttrFace("name", fsize=fsize, fgcolor=fgcolor)
+
+        faces.add_face_to_node(N, node, 0)
+    if "weight" in node.features:
+        # Creates a sphere face whose size is proportional to node's
+        # feature "weight"
+        if (isinstance(node.weight, int) or isinstance(node.weight, float)):
+            weight = [node.weight]
+        else:
+            weight = node.weight
+        C = BarChartFace(values=weight, width=width, height=height,
+                         colors=['#0000FF'], min_value=min_value,
+                         max_value=max_value)
+        # Let's make the sphere transparent
+        C.opacity = 0.5
+        # Rotate the faces by 270*
+        C.rotation = 270
+        # And place as a float face over the tree
+        faces.add_face_to_node(C, node, 0, position="float")
+
+
+def balanceplot(balances, tree,
+                layout=None,
+                mode='c'):
+    """ Plots balances on tree.
+
+    Parameters
+    ----------
+    balances : np.array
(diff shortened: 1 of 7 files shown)


antgonza:
I think you can test using [ete's get_ascii]( http://etetoolkit.org/docs/latest/reference/reference_tree.html#ete3.TreeNode.get_ascii), have you tried? 

BTW ipython notebooks render correctly and you can comment on review, for example: [Platypus](https://github.com/biocore/Platypus-Conquistador/blob/master/ipynbs/Chasing_Platypus_via_MG-RAST.ipynb)

ElDeveloper:
This is slightly off-topic, but just in case this comes up, the rendering by GitHub has some limitations for execution of custom JavaScript code within Jupyter notebooks. However, nbviewer does not have any limitations, if you see [this link](http://nbviewer.jupyter.org/github/biocore/Platypus-Conquistador/blob/master/ipynbs/Chasing_Platypus_via_MG-RAST.ipynb), you'll see the correct rendering of the maps at the bottom of the notebook, while [here](https://github.com/biocore/Platypus-Conquistador/blob/master/ipynbs/Chasing_Platypus_via_MG-RAST.ipynb) you'll see no renders.

mortonjt:
Interesting - we won't be touching javascript (I *think*).  ETE uses PyQT4 in the backend.

Just added some tests for testing some of the functionality.  Should be good for another round of review once this passes.

mortonjt:
This is ready for another round of review

gneiss/balances.py
```diff
@@ -129,3 +131,163 @@ def _count_matrix(treenode):
             counts[n]['k'] = counts[n.parent]['k'] + counts[n.parent]['r']
             counts[n]['t'] = counts[n.parent]['t']
     return counts, n_tips
+
+
+def default_layout(node):
+    if node.is_leaf():
```
antgonza: Could you add a docstring?

gneiss/balances.py
```diff
@@ -129,3 +131,163 @@ def _count_matrix(treenode):
...
+        faces.add_face_to_node(C, node, 0, position="float")
+
+
+def barchart_layout(node, name='name',
+                    width=20, height=40,
+                    colors=None, min_value=0, max_value=1,
+                    fsize=14, fgcolor="black",
+                    alpha=0.5,
+                    rotation=270):
+    if colors is None:
```
antgonza: Could you add a docstring?

gneiss/balances.py
antgonza: This is not specifically tested, is there a reason for this?

antgonza:
A few comments, could you add coveralls? Thanks.

gneiss/balances.py
mortonjt: Yes.  Because I have no idea how to test it.  

I think this scenario is ok.  ETE layouts are only used to specify the parameters for the TreeStyle.

mortonjt:
Ok, I think this should be good.  I'm not sure how to get the badges working.  I'll fix them in a bit.

In the meantime, this is ready for a final round of review before merging.

antgonza:
The coverage is pretty bad 79.63% due to the missing tests. Could you add tests making sure that what you change in ETE's faces, actually changes? Other option is to separate those functions into another file that is not accounted for coverage ... I think the first option is better.

mortonjt:
I don't think we should be testing the layouts - I'm not defining some default layouts that can be used if user is lazy.

These layouts are designed to be user defined - there no tests in ETE to test these layouts.

ElDeveloper:
Could things be tested "ala matplotlib"? i.e. creating and saving a PNG
file? Or alternatively checking for specific features of the returned
objects?

wasade:
Matplotlib figure testing is not trivial. Good write up in AOSA
Could things be tested "ala matplotlib"? i.e. creating and saving a PNG
file? Or alternatively checking for specific features of the returned
objects?

On (Jul-06-16|11:49), Antonio Gonzalez wrote:
> The coverage is pretty bad 79.63% due to the missing tests. Could you add
tests making sure that what you change in ETE's faces, actually changes?
Other option is to separate those functions into another file that is not
accounted for coverage ... I think the first option is better.
>
> ---
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/biocore/gneiss/pull/4#issuecomment-230868280

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/biocore/gneiss/pull/4#issuecomment-230871278>, or mute
the thread
<https://github.com/notifications/unsubscribe/AAc8ssOVpBCw2UlOlhAqEK1bHHWKRnEiks5qS_sxgaJpZM4JEnCh>
.

antgonza:
Thanks, @mortonjt. Could you add an issue to add tests to the layouts?

mortonjt:
Just raised on issue.  Thanks!  Is this ready to go otherwise?

mortonjt:
Thanks for the all of the reviews!! They have been tremendously helpful for improving this code.

PR #6, opened 2016-07-08
ENH: Converting all calculates to assume level ordering

mortonjt:
As I was putting together the IPython notebooks, I realized that the basis was being returned in reverse level-ordering.  That is now being corrected here to make these modules a little less confusing.


gneiss/balances.py
```diff
@@ -169,7 +172,7 @@ def _attach_balances(balances, tree):
                 # The balances are oriented backwards
                 # so we need to rearrange them so that the
                 # ordering is correct.
-                n.add_features(weight=balances[-i])
+                n.add_features(weight=balances[i])
```
antgonza: Just to confirm, the comment above still applies, right?

mortonjt: Nope.  Good catch!

PR #8, opened 2016-07-11
PERF: more direct use of numpy

wasade:
@mortonjt, was thinking something along these lines. Note, the large test is failing and I'm not sure why. One issue I encountered was that I had to switch up the calculation of `k` and `t` (see [here](https://github.com/biocore/gneiss/compare/master...wasade:perf_tweaks?expand=1#diff-50b1a1692a03c26a330c3689f064d40aR143) as the original formulation was leading to more test failures. However, I fully recognize that that is probably why the large test is failing... anyway, any insight here would be great. I haven't benched this but I strongly suspect the `_balance_basis` call will be much more efficient, and there is still additional fat to trim if desired. 

Also, because of the failure, I think the print statements in the test code are fine as it may highlight a little more about what might be going on.

### DIFF
--- a/gneiss/balances.py
+++ b/gneiss/balances.py
@@ -10,25 +10,37 @@ from gneiss.layouts import default_layout
 def _balance_basis(tree_node):
     """ Helper method for calculating balance basis
     """
-    counts, n_tips = _count_matrix(tree_node)
-    counts = OrderedDict([(x, counts[x])
-                          for x in counts.keys() if not x.is_tip()])
-    nds = counts.keys()
-    r = np.array([counts[n]['r'] for n in nds])
-    s = np.array([counts[n]['l'] for n in nds])
-    k = np.array([counts[n]['k'] for n in nds])
-    t = np.array([counts[n]['t'] for n in nds])
+    # TODO: use recarray
+    # col 0 -> right counts
+    # col 1 -> left counts
+    # col 2 -> k
+    # col 3 -> t
+    r_idx = 0
+    l_idx = 1
+    k_idx = 2
+    t_idx = 3
+
+    counts, n_tips, n_nodes = _count_matrix(tree_node)
+    r = counts[:, r_idx]
+    s = counts[:, l_idx]
+    k = counts[:, k_idx]
+    t = counts[:, t_idx]
 
     a = np.sqrt(s / (r*(r+s)))
     b = -1*np.sqrt(r / (s*(r+s)))
 
     basis = np.zeros((n_tips-1, n_tips))
-    for i in range(len(nds)):
-        basis[i, :] = np.array([0]*k[i] + [a[i]]*r[i] + [b[i]]*s[i] + [0]*t[i])
-    # Make sure that the basis is in level order
-    basis = basis[:, ::-1]
-    nds = list(nds)
-    return basis, nds
+    for i in np.arange(n_nodes - n_tips, dtype=int):
+        v = basis[i]
+
+        k_i = n_tips - k[i]
+        r_i = k_i - r[i]
+        s_i = r_i - s[i]
+
+        v[r_i:k_i] = a[i]
+        v[s_i:r_i] = b[i]
+
+    return basis, [n for n in tree_node.levelorder() if not n.is_tip()]
 
 
 def balance_basis(tree_node):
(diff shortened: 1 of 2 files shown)


mortonjt:
This is looking awesome!!  Thanks for putting this together!

Performance is definitely worth considering - but given that rendering the trees takes much more time than the actual calculations, it isn't the highest priority.

However, I think the main strength of this PR comes from replacing the TreeNode objects with numpy arrays, which is a much more robust approach around this problem.

What do you think about stripping some of this functionality out to create a `BalanceMatrix`  or a  `BalanceTree` object designed for constructing and querying balances?  One of the biggest issues now is actually querying, manipulating and interpreting the results.  We can even use this opportunity as a testing ground for the contingency table.

gneiss/balances.py
```diff
@@ -90,51 +102,63 @@ def balance_basis(tree_node):
...
+            try:
+                left, right = n.children
+            except:
+                raise ValueError("Not a strictly bifurcating tree!")
+            n._tip_count = left._tip_count + right._tip_count
+
+    # TODO: use recarray
+    # col 0 -> right counts
+    # col 1 -> left counts
+    # col 2 -> k
```
mortonjt: To clarify, k is the number of tips left of the subtree.

So if we have the following tree

```

                                        /-a
                              /5-------|
                             |          \-b
                    /3-------|
                   |         |          /-f
                   |          \6-------|
          /2-------|                    \-d
         |         |
         |         |          /-h
-1-------|          \4-------|
         |                    \-j
         |
          \-l
```
And we look at node 4
`k=1` since there is only 1 tip right of the subtree rooted at `4`
`t=4` since there are 4 tips left of the subtree rooted at `4`

gneiss/balances.py
```diff
@@ -90,51 +102,63 @@ def balance_basis(tree_node):
...
         else:
-            counts[n]['k'] = counts[n.parent]['k'] + counts[n.parent]['r']
-            counts[n]['t'] = counts[n.parent]['t']
-    return counts, n_tips
+            parent_counts = counts[n.parent._lo_idx]
+            if n is n.parent.children[0]:
+                #t = parent_counts[t_idx] + parent_counts[l_idx]
+                #k = parent_counts[k_idx]
+
+                k = parent_counts[k_idx] + parent_counts[r_idx]
```
mortonjt: This looks a little backwards.  Stepping through the previous example I gave before

```
                                       /-a
                              /5-------|
                             |          \-b
                    /3-------|
                   |         |          /-f
                   |          \6-------|
          /2-------|                    \-d
         |         |
         |         |          /-h
-1-------|          \4-------|
         |                    \-j
         |
          \-l
```
If we were trying to calculate `k` and `t` for balance `4`, and these values were all properly calculated for balance `2`, the following will happen.

```python
k = parent_counts[k_idx] + parent_counts[r_idx]
  = 0 + 2 = 2
t = parent_counts[t_idx]
  = 0
```
When we know that `k=1` and `t=4` for the balance `4`.  Does this make more sense?

gneiss/balances.py
```diff
@@ -90,51 +102,63 @@ def balance_basis(tree_node):
...
+
+        if n.is_root():
+            k = 0
+            t = 0
         else:
-            counts[n]['k'] = counts[n.parent]['k'] + counts[n.parent]['r']
-            counts[n]['t'] = counts[n.parent]['t']
-    return counts, n_tips
+            parent_counts = counts[n.parent._lo_idx]
+            if n is n.parent.children[0]:
```
mortonjt: One thing that may not be obvious from this line of code is - this is supposed to determine if the current node is on the left side of the parent node, or the right side of the parent node.

wasade: Yes, that's what it does

gneiss/balances.py
wasade: Tests don't pass if flipped though hence the confusion

gneiss/balances.py
mortonjt: Ok cool.  Just wanted to double check to make sure we are on the same page.

gneiss/balances.py
wasade: Given that example, it sounds like k is concerned with immediate
descendents and t is concerned with the descendents of its sibling? Or do
you mean to consider node 2?

wasade:
What do the queries look like that are being difficult? Unclear to me if
doing a custom data type makes sense without knowing more about the desired
operations

mortonjt:
I like this tree, so I'll keep using it.
```
                                        /-a
                              /-------|
                             |          \-b
                    /-------|
                   |         |          /-f
                   |          \-------|
          /-------|                    \-d
         |         |
         |         |          /-h
--------|          \-------|
         |                    \-j
         |
          \-l
```
Say we have the tree, where none of the internal nodes are labeled. This is the case with most of the trees that I'm dealing with.  Even if tax2tree is functional, this is still a problem, since we are also constructing trees from UPGMA.

Right now, the way that I'm dealing with this problem is assigning level ordering to each of the node names.  However, there are a few problems with this

1. It will overwrite any existing naming scheme for the internal nodes
2. Its a ton of boiler plate code
3. Its very difficult to interpret - requires extensive querying + visualization of the tips.

This is definitely out of the scope of this PR.  Thinking about raising an issue about this.

wasade:
If you aren't concerned about order, suggest just using assign_ids and use
the .id attribute. Or just decorate on an attribute as is done in this pr.
A lookup dict can be maintained keyed by id. If this is the only type of
queries then recommend against a subclass or custom data type

I don't understand the tax2tree comment, can you clarify?

Agree re: separate issue.

gneiss/balances.py
mortonjt: `k`  is more concerned about the tips to the right of `4` and `t` is more concerned about the tips left of `4`.  I realize that the orientation is a little confusing.

The reason why we want to do this, is because at each internal node of the tree, all of the tips of the tree fall into 2 categories.
Tips to the right of the node, but are not contained in the subtree (`k`)
Tips that are descendants of the node, but are in the right subtree (`r`)
Tips that are descendants of the node, but are in the left subtree (`l`)
Tips to the left of the node, but are not contained in the subtree (`t`)

So these 4 numbers should always add up to
`k+r+l+t = N` where `N` is the total number of tips in the tree.

wasade: Ah, okay, thanks. So in the tree as a whole, this makes much more sense.
Will kick this around more tomorrow

mortonjt: Thanks!

mortonjt:
Thanks for the PR @wasade !  This helped me wrap my head around how to create a sparse version of this.  Closing this in favor of #250

PR #12, opened 2016-07-19
Util

mortonjt:
This is adding in the utility helper functions discussed in code review.

@antgonza @ElDeveloper @josenavas, mind if you could sanity check this PR?  Thanks!!

Note that this depends on #13

### DIFF
--- a/gneiss/layouts.py
+++ b/gneiss/layouts.py
@@ -1,3 +1,11 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, gneiss development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
 from ete3 import faces, AttrFace, CircleFace, BarChartFace
 
 
--- a/gneiss/tests/test_util.py
+++ b/gneiss/tests/test_util.py
@@ -1,0 +1,241 @@
+import unittest
+import pandas as pd
+import pandas.util.testing as pdt
+from skbio import TreeNode
+from gneiss.util import match, match_tips, rename_tips
+
+
+class TestUtil(unittest.TestCase):
+
+    def test_match(self):
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [2, 2, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['o1', 'o2', 'o3', 'o4'])
+        metadata = pd.DataFrame([['a', 'control'],
+                                 ['b', 'control'],
+                                 ['c', 'diseased'],
+                                 ['d', 'diseased']],
+                                index=['s1', 's2', 's3', 's4'],
+                                columns=['Barcode', 'Treatment'])
+        exp_table, exp_metadata = table, metadata
+        res_table, res_metadata = match(table, metadata)
+        pdt.assert_frame_equal(exp_table, res_table)
+        pdt.assert_frame_equal(exp_metadata, res_metadata)
+
+
+    def test_match_scrambled(self):
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [2, 2, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['o1', 'o2', 'o3', 'o4'])
+        metadata = pd.DataFrame([['a', 'control'],
+                                 ['c', 'diseased'],
+                                 ['b', 'control'],
+                                 ['d', 'diseased']],
+                                index=['s1', 's3', 's2', 's4'],
+                                columns=['Barcode', 'Treatment'])
+        exp_table = table
+        exp_metadata =  pd.DataFrame([['a', 'control'],
+                                      ['b', 'control'],
+                                      ['c', 'diseased'],
+                                      ['d', 'diseased']],
+                                     index=['s1', 's2', 's3', 's4'],
+                                     columns=['Barcode', 'Treatment'])
+
+        res_table, res_metadata = match(table, metadata)
+        pdt.assert_frame_equal(exp_table, res_table)
+        pdt.assert_frame_equal(exp_metadata, res_metadata)
+
+    def test_match_intersect(self):
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [2, 2, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['o1', 'o2', 'o3', 'o4'])
+        metadata = pd.DataFrame([['a', 'control'],
+                                 ['c', 'diseased'],
+                                 ['b', 'control']],
+                                index=['s1', 's3', 's2'],
+                                columns=['Barcode', 'Treatment'])
+
+        exp_table = pd.DataFrame([[0, 0, 1, 1],
+                                  [2, 2, 4, 4],
+                                  [5, 5, 3, 3]],
+                                 index=['s1', 's2', 's3'],
+                                 columns=['o1', 'o2', 'o3', 'o4'])
+
+        exp_metadata =  pd.DataFrame([['a', 'control'],
+                                      ['b', 'control'],
+                                      ['c', 'diseased']],
+                                     index=['s1', 's2', 's3'],
+                                     columns=['Barcode', 'Treatment'])
+
+        res_table, res_metadata = match(table, metadata, intersect=True)
+        pdt.assert_frame_equal(exp_table, res_table)
+        pdt.assert_frame_equal(exp_metadata, res_metadata)
+
+    def test_match_mismatch(self):
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [2, 2, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['o1', 'o2', 'o3', 'o4'])
+        metadata = pd.DataFrame([['a', 'control'],
+                                 ['c', 'diseased'],
+                                 ['b', 'control']],
+                                index=['s1', 's3', 's2'],
+                                columns=['Barcode', 'Treatment'])
+
+        exp_table = pd.DataFrame([[0, 0, 1, 1],
+                                  [2, 2, 4, 4],
+                                  [5, 5, 3, 3]],
+                                 index=['s1', 's2', 's3'],
+                                 columns=['o1', 'o2', 'o3', 'o4'])
+
+        exp_metadata =  pd.DataFrame([['a', 'control'],
+                                      ['b', 'control'],
+                                      ['c', 'diseased']],
+                                     index=['s1', 's2', 's3'],
+                                     columns=['Barcode', 'Treatment'])
+        with self.assertRaises(ValueError):
+            match(table, metadata)
+
+    def test_match_tips(self):
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [2, 2, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['a', 'b', 'c', 'd'])
+        tree = TreeNode.read([u"(((a,b)f, c),d)r;"])
+        exp_table, exp_tree = table, tree
+        res_table, res_tree = match_tips(table, tree)
+        pdt.assert_frame_equal(exp_table, res_table)
+        self.assertEqual(str(exp_tree), str(res_tree))
+
+    def test_match_tips_scrambled_tips(self):
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [2, 3, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['a', 'b', 'c', 'd'])
+        tree = TreeNode.read([u"(((b,a)f, c),d)r;"])
+        exp_tree = tree
+        exp_table = pd.DataFrame([[0, 0, 1, 1],
+                                  [3, 2, 4, 4],
+                                  [5, 5, 3, 3],
+                                  [0, 0, 0, 1]],
+                                 index=['s1', 's2', 's3', 's4'],
+                                 columns=['b', 'a', 'c', 'd'])
+
+        res_table, res_tree = match_tips(table, tree)
+        pdt.assert_frame_equal(exp_table, res_table)
+        self.assertEqual(str(exp_tree), str(res_tree))
+
+    def test_match_tips_scrambled_columns(self):
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [3, 2, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['b', 'a', 'c', 'd'])
+        tree = TreeNode.read([u"(((a,b)f, c),d)r;"])
+        exp_tree = tree
+        exp_table = pd.DataFrame([[0, 0, 1, 1],
+                                  [2, 3, 4, 4],
+                                  [5, 5, 3, 3],
+                                  [0, 0, 0, 1]],
+                                 index=['s1', 's2', 's3', 's4'],
+                                 columns=['a', 'b', 'c', 'd'])
+
+        res_table, res_tree = match_tips(table, tree)
+        pdt.assert_frame_equal(exp_table, res_table)
+        self.assertEqual(str(exp_tree), str(res_tree))
+
+
+    def test_match_tips_intersect_tips(self):
+        # there are less tree tips than table columns
+        table = pd.DataFrame([[0, 0, 1, 1],
+                              [2, 3, 4, 4],
+                              [5, 5, 3, 3],
+                              [0, 0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['a', 'b', 'c', 'd'])
+        tree = TreeNode.read([u"((a,b)f,d)r;"])
+        exp_table = pd.DataFrame([[0, 0, 1],
+                                  [2, 3, 4],
+                                  [5, 5, 3],
+                                  [0, 0, 1]],
+                                 index=['s1', 's2', 's3', 's4'],
+                                 columns=['a', 'b', 'd'])
+        exp_tree = tree
+        res_table, res_tree = match_tips(table, tree, intersect=True)
+        pdt.assert_frame_equal(exp_table, res_table)
+        self.assertEqual(str(exp_tree), str(res_tree))
+
+    def test_match_tips_intersect_columns(self):
+        # table has less columns than tree tips
+        table = pd.DataFrame([[0, 0, 1],
+                              [2, 3, 4],
+                              [5, 5, 3],
+                              [0, 0, 1]],
+                             index=['s1', 's2', 's3', 's4'],
+                             columns=['a', 'b', 'd'])
+        tree = TreeNode.read([u"(((a,b)f, c),d)r;"])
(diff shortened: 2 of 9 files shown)


gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns
+    -------
+    _x : pd.DataFrame
+        Filtered dataframe
+    _y : pd.DataFrame
+        Filtered dataframe
+    """
+    _x = x.sort_index()
+    _y = y.sort_index()
+    if intersect:
```
antgonza: What about adding a test that set(_x.index) and _x.index are the same length?

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+        contingency table and the metadata table will returned.
+
+    Returns
+    -------
+    _x : pd.DataFrame
+        Filtered dataframe
+    _y : pd.DataFrame
+        Filtered dataframe
+    """
+    _x = x.sort_index()
```
antgonza: Do we really need them sort?

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+    _y = y.sort_index()
+    if intersect:
+        idx = set(_x.index) & set(_y.index)
+        idx = sorted(idx)
+        return _x.loc[idx], _y.loc[idx]
+    else:
```
antgonza: In which cases it makes sense to not intersect?

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Returns
+    -------
+    pd.DataFrame :
+        Subset of the original contingency table with the common features.
+    skbio.TreeNode :
+        Sub-tree with the common features.
+    """
+    tips = [x.name for x in tree.tips()]
+    common_tips = list(set(tips) & set(table.columns))
```
antgonza: why do you need it as list?

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+
+    if intersect:
```
antgonza: Same question, why when intersect makes sense to be False? Perhaps you can have another method that check sizes and you could use for both cases.

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns
+    -------
+    skbio.TreeNode
+       Tree with renamed internal nodes.
+    """
+    i = 0
+    for n in tree.levelorder():
+        if not n.is_tip():
+            if names is None:
+                n.name = 'y%i' % i
```
antgonza: why use y? Perhaps using tn (tip-name) or ttn (temp-tip-name) or something like that will be better.

antgonza:
A few comments.

gneiss/tests/test_util.py
```diff
@@ -0,0 +1,225 @@
+import unittest
```
josenavas: I realized that none of this files have the copyright notice on top. Should it be added?

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
+def match(x, y, intersect=False):
+    """ Sorts samples in metadata and contingency table in the same order.
+
+    Parameters
+    ----------
+    x : pd.DataFrame
```
josenavas: I noticed that `x` and `y` are both DataFrames but conceptually are two different things. I would recommend to rename the variable to `cont_table` and `metadata` or similar, but at least more meaningful that `x` and `y`

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+        Specifies if only the intersection of samples in the
+        contingency table and the metadata table will returned.
+
+    Returns
+    -------
+    _x : pd.DataFrame
+        Filtered dataframe
+    _y : pd.DataFrame
+        Filtered dataframe
+    """
```
josenavas: Missing `Raises` section

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Parameters
+    ----------
+    x : pd.DataFrame
+        Contingency table where samples correspond to rows and
+        features correspond to columns.
+    y: pd.DataFrame
+        Metadata table where samples correspond to rows and
+        explanatory metadata variables correspond to columns.
+    intersect : bool, optional
```
josenavas: Not sure if it is required by the numpy doc but I usually recommend to put the default value in the description of the parameter. If it is not required by the standard, feel free to ignore.

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+        Specifies if only the intersection of samples in the
+        contingency table and the metadata table will returned.
+
+    Returns
+    -------
+    _x : pd.DataFrame
+        Filtered dataframe
```
josenavas: Probably better to change to `Filtered contingency table` and `Filtered metadata table`?

gneiss/util.py
josenavas: The documentation indicates that the function sorts the samples in both dataframes. However, no sort is performed in this `else` section. Does this mean that the tests are incomplete?

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+                             "`x` has %d rows, `y` has %d rows.  "
+                             "Consider setting `intersect=True`." %
+                             (len(_x.index), len(_y.index)))
+        return _x, _y
+
+
+def match_tips(table, tree, intersect=False):
+    """ Returns the OTU table and tree with matched tips.
+
+    Sorts the columns of the OTU table to match the tips in
```
josenavas: Probably worth mentioning the order of the tips. There are multiple valid tip orderings in a tree. The function [`tips` from skbio](http://scikit-bio.org/docs/latest/generated/skbio.tree.TreeNode.tips.html#skbio.tree.TreeNode.tips) provides the tree tips in postorder traversal of the tree.

gneiss/util.py
josenavas: Also noted that you're providing the output names while in the function below you're not. Not sure which is the correct way, but worth normalizing to one of them.

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+    the tree.  If the tree is multi-furcating, then the
+    tree is reduced to a bifurcating tree by randomly inserting
```
josenavas: Probably better to point here to the [skbio `bifurcate`](http://scikit-bio.org/docs/latest/generated/skbio.tree.TreeNode.bifurcate.html#skbio.tree.TreeNode.bifurcate) documentation

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+        if len(tips) != len(table.columns):
+            raise ValueError("`table` and `tree` have incompatible sizes, "
+                             "`table` has %d columns, `tree` has %d tips.  "
+                             "Consider setting `intersect=True`." %
+                             (len(table.columns), len(tips)))
+
+        _table = table
+        _tree = tree
+
+    _tree.bifurcate()
```
josenavas: Just to make sure, calling `bifurcarte` in an already bifurcating tree doesn't change it, correct?

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+    _tree.prune()
+    sorted_features = [n.name for n in _tree.tips()]
+    _table = _table.reindex_axis(sorted_features, axis=1)
+    return _table, _tree
+
+
+def rename_tips(tree, names=None):
+    """ Names the tree tips according to level ordering.
+
+    The tree will be traversed from top-down, left to right.
```
josenavas: Can you rewrite to `The tree will be traversed in level order (i.e. top-down, left to right).` so we use common tree terminology? Thanks!

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+    If there `names` is not specified, the node with the smallest label (y0)
```
josenavas: If ~~there~~ `names`...

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+        Specifies if only the intersection of samples in the
+        contingency table and the tree will returned.
+
+    Returns
+    -------
+    pd.DataFrame :
+        Subset of the original contingency table with the common features.
+    skbio.TreeNode :
+        Sub-tree with the common features.
+    """
```
josenavas: Missing `Raises` section.

gneiss/util.py
```diff
@@ -0,0 +1,117 @@
...
+    skbio.TreeNode
+       Tree with renamed internal nodes.
+    """
+    i = 0
+    for n in tree.levelorder():
+        if not n.is_tip():
+            if names is None:
+                n.name = 'y%i' % i
+            else:
+                n.name = names[i]
```
josenavas: What if `len(names) !=` number of internal nodes?

josenavas:
Few additional comments. Thanks!

gneiss/util.py
mortonjt: I didn't even think about that.  We shouldn't allow for duplicate ids.  Good catch!

gneiss/util.py
mortonjt: Only if `intersect=False`.  That is now fixed.

gneiss/util.py
mortonjt: The sort here would be performed before the `if` clause.  Now, there are sorts within the `if` and `else` clauses.

We want the sort to occur if the samples in the metadata and table are scrambled.  We want the samples in the same order.  Does this make more sense?

gneiss/util.py
mortonjt: Because you cannot use sets to index DataFrames

```python
In [1]: import pandas as pd
In [2]: table = pd.DataFrame([[0, 0, 1, 1],
   ...:                       [2, 2, 4, 4],
   ...:                       [5, 5, 3, 3],
   ...:                       [0, 0, 0, 1]],
   ...:                       index=['s1', 's2', 's3', 's4'],
   ...:                       columns=['o1', 'o2', 'o3', 'o4'])

In [3]: x = set(['o1','o2'])

In [4: table[x]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-313a5d343876> in <module>()
----> 1 table[x]

/Users/mortonjt/miniconda3/envs/gneiss/lib/python3.5/site-packages/pandas/core/frame.py in __getitem__(self, key)
   1995             return self._getitem_multilevel(key)
   1996         else:
-> 1997             return self._getitem_column(key)
   1998 
   1999     def _getitem_column(self, key):

/Users/mortonjt/miniconda3/envs/gneiss/lib/python3.5/site-packages/pandas/core/frame.py in _getitem_column(self, key)
   2002         # get column
   2003         if self.columns.is_unique:
-> 2004             return self._get_item_cache(key)
   2005 
   2006         # duplicate columns & possible reduce dimensionality

/Users/mortonjt/miniconda3/envs/gneiss/lib/python3.5/site-packages/pandas/core/generic.py in _get_item_cache(self, item)
   1346         """Return the cached item, item represents a label indexer."""
   1347         cache = self._item_cache
-> 1348         res = cache.get(item)
   1349         if res is None:
   1350             values = self._data.get(item)

TypeError: unhashable type: 'set'
```

gneiss/util.py
mortonjt: If you already have a tree with the exact OTUs as the table, and you just want to sort the OTUs in the table.  It would be nice if we could have an `intersect` parameter, just to make sanity checking much easier (i.e. OTUs were dropped accidentally).

gneiss/util.py
mortonjt: This is directly copied from my notebook.  In my particular use case, I used `y` because I was using the balances as a response.  I'm by no means married to it.  But I think `tn` or `ttn` would be equally confusing, since they aren't tip names, they are names for internal nodes.

What do you think about renaming this to `I`?

gneiss/tests/test_util.py
mortonjt: 👍

gneiss/util.py
mortonjt: Fixed!

gneiss/util.py
mortonjt: Done!

gneiss/util.py
mortonjt: Nah.  This is important to note.  Just added!

gneiss/util.py
mortonjt: Yup.  Just fixed.

gneiss/util.py
mortonjt: Noted!

gneiss/util.py
mortonjt: Done

gneiss/util.py
mortonjt: I hope not.  When I run this on a small example, it doesn't
```python
In [12]: from skbio import TreeNode

In [13]: tree = TreeNode.read([u"((a,b)f,d)r;"])

In [14]: tree
Out[14]: <TreeNode, name: r, internal node count: 1, tips count: 3>

In [15]: print(tree.ascii_art())
                    /-a
          /f-------|
-r-------|          \-b
         |
          \-d

In [16]: tree.bifurcate()

In [17]: print(tree.ascii_art())
                    /-a
          /f-------|
-r-------|          \-b
         |
          \-d

```

gneiss/util.py
mortonjt: Fixed.

gneiss/util.py
mortonjt: IndexError?  Probably should add a test for that.

gneiss/util.py
mortonjt: Done

gneiss/layouts.py
```diff
@@ -1,3 +1,11 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, gneiss development team.
+#
+# Distributed under the terms of the Modified BSD License.
```
wasade: until ete is BSD compatible, it is not appropriate to import ete3 objects into a BSD project...

gneiss/util.py
wasade: That would imply a serious bug in the code...

gneiss/util.py
wasade: :+1: for `y` or `i`. it may make sense to ensure the name does not collide with any other name in the tree. This loop ensures that it will not collide with internal nodes, but it is possible (though unlikely) to collide with a tip name. Since the decoration of a name is used for lookup purposes, it might be worth enforcing this invariant.

gneiss/util.py
josenavas: You can probably do a check and raise a more useful error? Just an index error will not help the developer using this function to debug.

mortonjt:
This PR is dependent on #13. This project needs to be GPLed before we add any more code in.

gneiss/layouts.py
mortonjt: Addressed in #13. Thanks!

gneiss/util.py
mortonjt: Just added a `ValueError`.   Let me know what you think!

gneiss/util.py
mortonjt: Done

mortonjt:
This is ready for review/merge.

gneiss/util.py
```diff
@@ -0,0 +1,172 @@
...
+    subtableids = set(table.index)
+    submetadataids = set(metadata.index)
+    if len(subtableids) != len(table.index):
+        raise ValueError("`table` has duplicate sample ids.")
+    if len(submetadataids) != len(metadata.index):
+        raise ValueError("`metadata` has duplicate sample ids.")
+
+    if intersect:
+        idx = subtableids & submetadataids
+        idx = sorted(idx)
```
antgonza: Why do we need the ids to be sorted?

gneiss/util.py
```diff
@@ -0,0 +1,172 @@
...
+        return table.loc[idx], metadata.loc[idx]
+    else:
+        subtable = table.sort_index()
+        submetadata = metadata.sort_index()
+
+        if len(subtable.index) != len(submetadata.index):
```
antgonza: Why do we need the table/metadata to be the same size? I thought the idea was to make them the same size ....

gneiss/util.py
antgonza: That's exactly my point. Wouldn't it make more sense to have a single method that receives two (or more?) parameters, checks they are the same size and returns the sorted index? This will simplify both match* methods by removing that functionality from them ...

gneiss/util.py
antgonza: Should we add a test for this or can we assume that this is a tests in skbio.TreeNode?

gneiss/util.py
mortonjt: Because otherwise the pandas dataframe will be scrambled due to the way that sets work.

Sorting them the only way to resolve this issue, for the sake of testing.

gneiss/util.py
mortonjt: Because the intersect operation is `False`.  We want the user to be completely aware that they are dropping samples.

gneiss/util.py
josenavas: @mortonjt if the sorting is just for testing, can you sort them in the test prior doing the assert_frame_equals? Sorting is O(nlogn) so you can avoid this in big datasets unless strictly needed. Does that make sense?

gneiss/util.py
josenavas: We shouldn't be adding a test for that in here, since that is not code generated here.

gneiss/util.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Returns
+    -------
+    skbio.TreeNode
+       Tree with renamed internal nodes.
+
+    ValueError:
+        Raised if `tree` and `name` have incompatible sizes.
+    """
+    _tree = tree.copy()
```
josenavas: Is the copy strictly necessary? I'm thinking if it will be okay to just do the renaming inplace - just thinking on the memory footprint of this operations as the tree size increases.

josenavas:
Thanks @mortonjt a couple of comments!

gneiss/util.py
mortonjt: I just saw this comment.  Still very unclear about this.

gneiss/util.py
mortonjt: Just removed the sort and moved over this over to the unittests.

gneiss/util.py
mortonjt: The copy isn't _strictly_ necessary.  But it is certainly useful.  Just added an `inplace` parameter here.  

It'll be tricky to have this functionality in the other modules, mainly because pandas and skbio don't make filter and `shear` operations in place (as far as I'm aware).  If you think this is necessary, we can create an issue for this, so that this PR can proceed.

gneiss/util.py
josenavas: I think what @antgonza is suggesting is trying to create a util function to perform this intersect operation and the one in the previous function, given that the 2 codes are pretty similar. However, I think that since the object types are different, you are not able to create that function. Am I right?

mortonjt: Exactly.  The types are a bit different, which is why I didn't do this before.

antgonza: I still believe that will reduce duplication of code. What I was thinking is something like this:
```python
def validate(vals1, vals2):
    if len(vals1) != len(vals2):
        raise ValueError(...)

validate(_x.index, _y.index)
validate(tips, table.columns)
```

What do you think?

josenavas: Sounds good, although I would advocate for a different name: `assert_length` or something like that.

antgonza: Yes, @mortonjt use whatever name you think is better.

gneiss/util.py
josenavas: I don't think I follow with the second part of your comment. With the `inplace` parameter my comment is resolved.

mortonjt: Sweet.  Then that settles it :)

gneiss/util.py
mortonjt: I still disagree that we should have this.

1. Its reducing the code from 4 lines to 2.  The only thing that is changing is the error message is being moved to an auxiliary function
2. This will actually make the error message + code less readable.  Since these are different data types - the information about the data type will need to be removed from the error message.

antgonza: Note that this will also remove an unnecessary parameter intersect in both methods and make it clear when you are validating the length of your inputs or trimming your sets. With this in mind and summarizing, it will reduce the number of parameters, the docstrings lengths, and the code in both methods. Thus, why I feel strongly about it.

mortonjt:
All comments addressed.  Should be good to go after tests pass.

PR #16, opened 2016-07-25
Nichesort

mortonjt:
Depends on #12.

Used to generate band tables when the gradient is known.

### DIFF
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -22,7 +22,7 @@ def mean_niche_estimator(abundances, gradient):
 
     .. math::
         E[g | x] =
-        \frac{1}{\sum\limits_{j=1}^N x_j} \sum\limits_{i=1}^N x_i g_i
+         \sum\limits_{i=1}^N x_i g_i \frac{x_i}{\sum\limits_{j=1}^N x_j}
 
     Where :math:`N` is the number of samples, :math:`x_i` is the proportion of
     species :math:`x` in sample :math:`i`, :math:`g_i` is the gradient value
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -66,7 +66,6 @@ def nichesort(table, gradient, niche_estimator=None):
     if niche_estimator is None:
         niche_estimator = mean_niche_estimator
 
-
     _table, _gradient = match(table, gradient)
 
     niche_estimator = partial(niche_estimator,
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -22,7 +22,7 @@ def mean_niche_estimator(abundances, gradient):
 
     .. math::
         E[g | x] =
-         \sum\limits_{i=1}^N x_i g_i \frac{x_i}{\sum\limits_{j=1}^N x_j}
+         \sum\limits_{i=1}^N g_i \frac{x_i}{\sum\limits_{j=1}^N x_j}
 
     Where :math:`N` is the number of samples, :math:`x_i` is the proportion of
     species :math:`x` in sample :math:`i`, :math:`g_i` is the gradient value
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -89,6 +89,8 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
         Sorted table according to the gradient of the samples, and the niches
         of the organisms along that gradient.
     """
+    if not callable(niche_estimator):
+        raise ValueError("`niche_estimator` is not a function.")
 
     table, gradient = match(table, gradient)
 
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -34,7 +34,8 @@ def mean_niche_estimator(abundances, gradient):
     # normalizes the proportions of the organism across all of the
     # samples to add to 1.
     v = abundances / abundances.sum()
-    return np.dot(gradient, v)
+    m = np.dot(gradient, v)
+    return m
 
 
 def nichesort(table, gradient, niche_estimator=None):
@@ -65,13 +66,14 @@ def nichesort(table, gradient, niche_estimator=None):
     if niche_estimator is None:
         niche_estimator = mean_niche_estimator
 
-    niche_estimator = partial(niche_estimator,
-                              gradient=gradient)
 
-    _table, _gradient = match(table, gradient, intersect=True)
+    _table, _gradient = match(table, gradient)
+
+    niche_estimator = partial(niche_estimator,
+                              gradient=_gradient)
+
     norm_table = _table.apply(lambda x: x/x.sum(), axis=1)
     est_niche = norm_table.apply(niche_estimator, axis=0)
-
     _gradient = _gradient.sort_values()
     est_niche = est_niche.sort_values()
 
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -1,3 +1,10 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, gneiss development team.
+#
+# Distributed under the terms of the GPLv3 License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
 import numpy as np
 import pandas as pd
 from functools import partial
@@ -7,9 +14,19 @@ from gneiss.util import match
 def mean_niche_estimator(abundances, gradient):
     """ Estimates the mean niche of an organism.
 
-    Calculates the mean niche or an organism along a gradient.
+    Calculates the mean niche of an organism along a gradient.
     This is done by calculating the expected value of an organism
     across the gradient.
+
+    Specifically, this module calculates the following
+
+    .. math::
+        E[g | x] =
+        \frac{1}{\sum\limits_{j=1}^N x_j} \sum\limits_{i=1}^N x_i g_i
+
+    Where :math:`N` is the number of samples, :math:`x_i` is the proportion of
+    species :math:`x` in sample :math:`i`, :math:`g_i` is the gradient value
+    at sample `i`.
 
     Parameters
     ----------
@@ -23,13 +40,22 @@ def mean_niche_estimator(abundances, gradient):
     -------
     np.float :
         The mean gradient that the organism lives in.
+
+    Raises
+    ------
+    ValueError:
+        If the length of `abundances` is not the same length as `gradient`.
+    ValueError:
+        If the length of `gradient` contains nans.
     """
-    if len(abundances) != len(gradient):
-        raise ValueError("Length of abundances (%d) doesn't match the length"
-                         " of the gradient (%d)" % (len(abundances),
-                                                    len(gradient)))
+    len_abundances = len(abundances)
+    len_gradient = len(gradient)
+    if len_abundances != len_gradient:
+        raise ValueError("Length of `abundances` (%d) doesn't match the length"
+                         " of the `gradient` (%d)" % (len_abundances,
+                                                      len_gradient))
     if np.any(pd.isnull(gradient)):
-        raise ValueError("`gradient` cannot have any missing values.")
+        raise ValueError("`gradient` cannot have any nans.")
 
     # normalizes the proportions of the organism across all of the
     # samples to add to 1.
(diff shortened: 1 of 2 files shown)


josenavas:
@mortonjt can you pull from upstream/master?

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
+import numpy as np
```
josenavas: Missing license notice.

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+import pandas as pd
+from functools import partial
+from gneiss.util import match
+
+
+def mean_niche_estimator(abundances, gradient):
+    """ Estimates the mean niche of an organism.
+
+    Calculates the mean niche or an organism along a gradient.
```
josenavas: `or` -> `of`?

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+        Vector of fraction abundances of an organism over
+        a list of samples.
+    gradient : pd.Series, np.float
+        Vector of numerical gradient values.
+
+    Returns
+    -------
+    np.float :
+        The mean gradient that the organism lives in.
+    """
```
josenavas: Missing raises section.

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+from functools import partial
+from gneiss.util import match
+
+
+def mean_niche_estimator(abundances, gradient):
+    """ Estimates the mean niche of an organism.
+
+    Calculates the mean niche or an organism along a gradient.
+    This is done by calculating the expected value of an organism
+    across the gradient.
```
josenavas: Since it is a simple mathematical function, can you add it in the documentation? I think in skbio they allow embedding latex-style syntax so it may be worth adding it here too.

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+        raise ValueError("`gradient` cannot have any missing values.")
+
+    # normalizes the proportions of the organism across all of the
+    # samples to add to 1.
+    v = abundances / abundances.sum()
+    m = np.dot(gradient, v)
+    return m
+
+
+def nichesort(table, gradient, niche_estimator=None):
```
josenavas: `nichesort` -> `niche_sort`

josenavas: Would make sense to do: 
```python
def niche_sort(table, gradient, niche_estimator= mean_niche_estimator):
````
The you can remove the if statement below

File: gneiss/sort.py [hunk 6]
```diff
@@ -0,0 +1,81 @@
...
+    and otus by their estimated niche along the gradient.
+
+    Parameters
+    ----------
+    table : pd.DataFrame
+        Contingency table where samples are rows and
+        features (i.e. OTUs) are columns.
+    gradient : pd.Series
+        Vector of numerical gradient values.
+    niche_estimator : function
```
josenavas: add `, optional` and a description of the default (ok, to just add the name and add the function in the `See Also` numpy doc section).

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if niche_estimator is None:
+        niche_estimator = mean_niche_estimator
+
+    _table, _gradient = match(table, gradient)
+
+    niche_estimator = partial(niche_estimator,
+                              gradient=_gradient)
+
+    norm_table = _table.apply(lambda x: x/x.sum(), axis=1)
```
josenavas: What is the difference between this normalization and the one done in the mean_niche_estimator function? It seems redundant, so I think you should be able to just do it in one of the functions (choose in which one makes more sense for your application).

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+    est_niche = norm_table.apply(niche_estimator, axis=0)
+    _gradient = _gradient.sort_values()
+    est_niche = est_niche.sort_values()
+
+    _table = _table.reindex(index=_gradient.index)
```
josenavas:
```python
_table = _table.reindex(index=_gradient.index, columns=est_niche.index)
```
and remove line below.

Or even:
```python
return _table.reindex(index=_gradient.index, columns=est_niche.index)
```

gneiss/tests/test_sort.py
```diff
@@ -0,0 +1,124 @@
+import numpy as np
```
josenavas: Missing license notice

gneiss/tests/test_sort.py
```diff
@@ -0,0 +1,124 @@
...
+            columns=['s1', 's2', 's3', 's4', 's5'],
+            index=['o1', 'o2', 'o3', 'o4']).T
+        gradient = pd.Series(
+            [1, 2, 3, 4, 5],
+            index=['s1', 's2', 's3', 's4', 's5'])
+        res_table = nichesort(table, gradient)
+        pdt.assert_frame_equal(table, res_table)
+
+    def test_basic_nichesort_scrambled(self):
+
```
josenavas: Do you mind adding a short comment saying that columns 1 and 2 as well rows 1 and 2 are the ones that are in the wrong order? It took me some time to figure out what was scrambled :-)

josenavas:
Some comments. Thanks @mortonjt

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+        a list of samples.
+    gradient : pd.Series, np.float
+        Vector of numerical gradient values.
+
+    Returns
+    -------
+    np.float :
+        The mean gradient that the organism lives in.
+    """
+    if len(abundances) != len(gradient):
```
antgonza: I don't feel strongly about this but normally we do something like this to minimize how many times you calculate the lenght of the iterables:
``` python
len_abundances = len(abundances)
len_gradient = len(gradient)
if len_abundances != len_gradient:
    ....
```

gneiss/sort.py
antgonza: Agree.

gneiss/sort.py
```diff
@@ -0,0 +1,81 @@
...
+    -------
+    pd.DataFrame :
+        Sorted table according to the gradient of the samples, and the niches
+        of the organisms along that gradient.
+    """
+
+    if niche_estimator is None:
+        niche_estimator = mean_niche_estimator
+
+    _table, _gradient = match(table, gradient)
```
antgonza: why do you create other variables?
`_table, _gradient = match(table, gradient)`
vs.
`table, gradient = match(table, gradient)`

antgonza:
some more comments.

gneiss/sort.py
mortonjt: Fixed

gneiss/sort.py
mortonjt: Done

gneiss/sort.py
mortonjt: At first thought, I thought this was obvious.

But perhaps its a little confusing in its current state - definitely worth enabling sphinx docs before the next release.  Raising an issue.

For now, the MathJax notation will be added, but not rendered.

gneiss/sort.py
mortonjt: Done

gneiss/sort.py
mortonjt: This normalization converts the raw counts into OTU proportions in each sample.  This is so that the total abundance of all OTUs within a sample sums to 1 (i.e. so that all of the OTUs are represented by probabilities).

Now, the `mean_niche_estimator` is doing something a little different.  It is calculating a [conditional expectation](https://en.wikipedia.org/wiki/Conditional_expectation)

More specifically, given the probability mass of an OTU, what is the expected gradient?  Since this is now a conditional expectation, the values of the gradient are weighted by the probability mass of a single OTU _across_ samples.  Not the original OTU proportions.

gneiss/sort.py
mortonjt: 👍

gneiss/tests/test_sort.py
mortonjt: Fixed

gneiss/tests/test_sort.py
mortonjt: Done

gneiss/sort.py
mortonjt: Done

gneiss/sort.py
mortonjt: Did it because I was worried about mutability.  Adding in an extra test addressing this.

mortonjt:
Thanks for taking the time to review this!  All comments have been addressed.

gneiss/sort.py
```diff
@@ -0,0 +1,109 @@
...
+        raise ValueError("Length of `abundances` (%d) doesn't match the length"
+                         " of the `gradient` (%d)" % (len_abundances,
+                                                      len_gradient))
+    if np.any(pd.isnull(gradient)):
+        raise ValueError("`gradient` cannot have any nans.")
+
+    # normalizes the proportions of the organism across all of the
+    # samples to add to 1.
+    v = abundances / abundances.sum()
+    m = np.dot(gradient, v)
```
antgonza: Is this line really doing this:
`E[g | x] = \frac{1}{\sum\limits_{j=1}^N x_j} \sum\limits_{i=1}^N x_i g_i`

mortonjt: Yup

gneiss/sort.py
```diff
@@ -0,0 +1,109 @@
...
+    return m
+
+
+def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
```
antgonza: should this method raise an error if niche_estimator is not defined? For example:
niche_sort(table, gradient, niche_estimator='jamie')

mortonjt: It should already raise one, since it isn't callable?

I guess I'm little confused about what suggesting.  I'm also not sure I could check for this.

antgonza: It does but perhaps it will be more helpful to give an error with the available methods vs. a simple not defined:
```python
In [1]: def estimator(a, b):
   ...:     return a+b
   ...: 

In [2]: def command(a, b, method=estimator):
   ...:     return method(a, b)
   ...: 

In [3]: command(1, 2)
Out[3]: 3

In [4]: command(1, 2, method=bla)
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-4-b39186e5fe69> in <module>()
----> 1 command(1, 2, method=bla)

NameError: name 'bla' is not defined
```

However, I can see that this is way beyond the scope of this PR, perhaps adding an issue will be OK for now, what do you think?

josenavas: @mortonjt this can be done easily:

```python
In [26]: def estimator(a, b):
   ....:     return a + b
   ....: 

In [27]: def command(a, b, method=estimator):
   ....:     if not callable(method):
   ....:         raise ValueError("Method is not a function!")
   ....:     return method(a, b)
   ....: 

In [28]: command(1, 2)
Out[28]: 3

In [29]: command(1, 2, method='foo')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-29-b322e668d915> in <module>()
----> 1 command(1, 2, method='foo')

<ipython-input-27-db5a10611032> in command(a, b, method)
      1 def command(a, b, method=estimator):
      2     if not callable(method):
----> 3         raise ValueError("Method is not a function!")
      4     return method(a, b)
      5 

ValueError: Method is not a function!
```

mortonjt: Done.  Thanks!

gneiss/sort.py
```diff
@@ -0,0 +1,111 @@
...
+        object. The ability for the object to be ordered is critical, since
+        this will allow the table to be sorted according to this ordering.
+        By default, `mean_niche_estimator` will be used.
+
+    Returns
+    -------
+    pd.DataFrame :
+        Sorted table according to the gradient of the samples, and the niches
+        of the organisms along that gradient.
+    """
```
josenavas: missing raises section

josenavas:
Thanks @mortonjt ! The comments really help to fully understand what is going on in the code! Just a minor comment about missing a section in the docs and I think this is good to go!

gneiss/sort.py
mortonjt: Doh.  Added

josenavas:
Good to merge - I just realized that I don't have merge privileges here

gneiss/sort.py
antgonza: @mortonjt and I discussed offline and realized that the code and the formula are similar but not the same, one was a simplification of the other. @mortonjt fixed it.

mortonjt:
Thanks for your insights on this!  This feedback is seriously improving the usability of this module.

PR #21, opened 2016-07-28
ENH: Adding in first draft of the RegressionResults object

mortonjt:


mortonjt:
@antgonza @josenavas @wasade this is ready for a first round of review :)

gneiss/_summary.py
```diff
@@ -0,0 +1,52 @@
...
+        """ Reorganizes statsmodels regression modules.
+
+        Accepts a list of statsmodels RegressionResults objects
+        and performs some addition summary statistics.
+
+        Parameters
+        ----------
+        stat_results : list, sm.RegressionResults
+            List of RegressionResults objects.
+        feature_names : array_like, str
```
antgonza: optional?

gneiss/_summary.py
```diff
@@ -0,0 +1,52 @@
...
+        self.feature_names = feature_names
+        self.basis = basis
+        self.results = stat_results
+
+        # obtain pvalues
+        pvals = pd.DataFrame()
+        for i in range(len(self.results)):
+            p = self.results[i].pvalues
+            p.name = self.results[i].model.endog_names
+            pvals = pvals.append(p)
```
antgonza: Are you sure this is right?
```python
In [1]: a = [1]

In [2]: a.append(2)

In [3]: a
Out[3]: [1, 2]

In [4]: a = a.append(3)

In [5]: a

In [6]: type(a)
Out[6]: NoneType
```

mortonjt: I hope so
```python
>>> import pandas as pd
>>> pvals = pd.DataFrame()
>>> pvals.append(pd.Series([1, 2, 3], name='X', index=['a', 'b', 'c']))
>>> pvals
     a    b    c
X  1.0  2.0  3.0
```

gneiss/_summary.py
mortonjt: yup

gneiss/_summary.py
antgonza: Got it, I was reading this as arrays and not as pandas DataFrame, thanks.

gneiss/_summary.py
```diff
@@ -0,0 +1,52 @@
...
+        self.pvalues = pvals
```
antgonza: Do you really need to create pvals? I can see replacing all the pvals for self.pvalues, right?

gneiss/_summary.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # calculate the overall R2 value
+        sse = sum([r.ssr for r in self.results])
```
antgonza: In this and the next line you are looping again self.results which you are looping over in the for loop, right? Perhaps more efficient:
```python
         sse = []
         ssr = []
         for i in range(len(self.results)):
             p = self.results[i].pvalues
             p.name = self.results[i].model.endog_names
             pvals = pvals.append(p)
             ssr.append(self.results[i].ssr)
             sse.append(self.results[i].ess)
```
BTW (1) why change the name from ess to sse? (2) will it be possible to loop over the elements vs the index of self.result?

mortonjt: Here's some timings.  Some boiler plate code is below so that you can reproduce this.

```python
In [1]: import pandas as pd
In [2]: from gneiss._summary import RegressionResults
In [3]: import statsmodels.formula.api as smf
In [6]: data = pd.DataFrame([[1, 3, 4, 5, 2, 3, 4],
   ...:                       list(range(1, 8)),
   ...:                       [1, 3, 2, 4, 3, 5, 4]],
   ...:                      columns=['s1', 's2', 's3', 's4',
   ...:                               's5', 's6', 's7'],
   ...:                      index=['Y1', 'Y2', 'X']).T
In [10]: model1 = smf.ols(formula="Y1 ~ X", data=data)
In [11]: model2 = smf.ols(formula="Y2 ~ X", data=data)
In [12]: results = [model1.fit(), model2.fit()]
```

When I run my code, here's the following time
```python
In [15]: %timeit RegressionResults(results)
100 loops, best of 3: 2.75 ms per loop
```

Your suggested needed a little tweaking (i.e. `ssr` and `sse` shouldn't be lists, they should be number).  I modified that code as follows

```python
        sse = 0
        ssr = 0
        for i in range(len(self.results)):
            p = self.results[i].pvalues
            p.name = self.results[i].model.endog_names
            pvals = pvals.append(p)
            ssr += self.results[i].ssr
            sse += self.results[i].ess
```
When I run that code, I get the following timings

```python
In [8]: %timeit RegressionResults(results)
100 loops, best of 3: 5.31 ms per loop
```
So about twice as slow.

What's probably going on is there is that the list comprehension + np.sum is vectorized, allowing for more compiler optimizations.  Plus it is much more readable IMHO.

Do you think this is still ok to keep as is?

mortonjt: Forgot to mention.

The reason why I'm using `ssr` instead of `ess` and `sse` instead of `ssr` is because I find the statsmodels naming convention really confusing.  In the statistical literature, `ssr` typically stands for "sum of squares regression", not "sum of squares residuals".  And `sse` typically stands for "sum of squares error".

I actually opened up an issue [here](https://github.com/statsmodels/statsmodels/issues/3131).  I'd prefer to keep those variables as is.  I'll add more documentation if necessary.

antgonza: Thanks!! I _think_ the issue is using i, could you test:
```python
        sse = 0
        ssr = 0
        for i in range(len(self.results)):
            r = self.results[i]
            p = r.pvalues
            p.name = r.model.endog_names
            pvals = pvals.append(p)
            ssr += r.ssr
            sse += r.ess
```
and 
```python
        sse = 0
        ssr = 0
        for r in self.results:
            p = r.pvalues
            p.name = r.model.endog_names
            pvals = pvals.append(p)
            ssr += r.ssr
            sse += r.ess
```

antgonza: naming: got it, thanks!

gneiss/_summary.py
mortonjt: Done

gneiss/_summary.py
mortonjt: :D

gneiss/_summary.py
mortonjt: Your method
```python
In [8]: %timeit RegressionResults(results)
100 loops, best of 3: 2.64 ms per loop
```
My method with your suggestion
```python
In [8]: %timeit RegressionResults(results)
100 loops, best of 3: 2.69 ms per loop
```
Loses by 0.05ms.  I'll put in your suggestions.  Thanks!

PR #20, opened 2016-07-28
Integration with statsmodels [DO NOT MERGE]

mortonjt:
Here's a large PR that I've been putting together to enable tight integration with statsmodels.
This PR will be split up into many smaller PRs, to lighten reviewer burden.   

Basically, this PR will be consist of the following mini PRs.
- [ ] OLS - Ordinary Least Squares module from statsmodels.  
- [ ] GLS - Generalized Least Squares module from statsmodels. 
- [ ] LME - Linear Mixed Effects Models  module from statsmodels. 
- [ ] GEE - Generalized Estimating Equations module from statsmodels. 
- [ ] RegressionResults - summarizes regression results on all balances (i.e. Ordinary Least Squares, Generalized Least Squares, ...)
- [ ] DifferientialAbundanceResults - summarizes differiential abundance results on balances (i.e. Linear Mixed Effects Models, Generalized Estimating Equations, ...)

Issue #11, opened 2016-07-16
Interface to statsmodels


mortonjt:
It would be amazing if we could take advantage of the formula interface in [statsmodels](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/formula/api.py) to run statistical tests on the individual balances.

mortonjt:
This has been knocked by #39

PR #22, opened 2016-07-28
Balance Regression

mortonjt:
#Addressing #11 and continuing from #20 

Here's a large PR that I've been putting together to enable tight integration with statsmodels.
This PR will be split up into many smaller PRs, to lighten reviewer burden.   

Basically, this PR will be consist of the following mini PRs.
- [x] OLS - Ordinary Least Squares module from statsmodels.  (#33)
- [x] LME - Linear Mixed Models from statsmodels.
- [x] RegressionResults - summarizes regression results on all balances (i.e. Ordinary Least Squares, Generalized Least Squares, ...) (#21 #28 #32)

### DIFF
--- a/gneiss/_formula.py
+++ b/gneiss/_formula.py
@@ -1,0 +1,98 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, gneiss development team.
+#
+# Distributed under the terms of the GPLv3 License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+import statsmodels.formula.api as smf
+from util import match, match_tips
+from gneiss._summary import RegressionResults
+
+
+def _process(table, metadata, tree):
+    """ Matches tips, features and samples between the table, metadata
+    and tree.  This module returns the features and samples that are
+    contained in all 3 objects.
+    """
+
+    _table, _metadata = match(table, metadata)
+    _table, _tree = match_tips(table, tree)
+    non_tips_no_name = [(n.name is None) for n in tree.levelorder() if n.is_tip()]
+
+    if any(non_tips_no_name):
+        _tree = rename_internal_nodes(_tree)
+    return _table, _metadata, _tree
+
+def ols(formula, table, metadata, tree, **kwargs):
+    """ Ordinary Least Squares applied to balances.
+
+    A ordinary least square regression is applied to each balance.
+
+    Parameters
+    ----------
+    formula : str
+        Formula representing the statistical equation to be evaluated.
+        These strings are similar to how equations are handled in R.
+        Note that the dependent variable in this string should not be
+        specified, since this method will be run on each of the individual
+        balances. See `patsy` for more details.
+    table : pd.DataFrame
+        Contingency table where samples correspond to rows and
+        features correspond to columns.
+    metadata: pd.DataFrame
+        Metadata table that contains information about the samples contained
+        in the `table` object.  Samples correspond to rows and covariates correspond
+        to columns.
+    tree : skbio.TreeNode
+        Tree object where the leaves correspond to the columns contained in
+        the table.
+    **kwargs : dict
+        Other arguments accepted into `statsmodels.regression.linear_model.OLS`
+
+    Returns
+    -------
+    RegressionResuls
+        Container object that holds information about the overall fit.
+
+    See Also
+    --------
+    statsmodels.regression.linear_model.OLS
+    """
+    _table, _metadata, _tree = _process(table, metadata, tree)
+    basis, _ = balance_basis(_table, _tree)
+
+    fits = []
+    for b in balance_names:
+        # mixed effects code is obtained here:
+        # http://stackoverflow.com/a/22439820/1167475
+        stats_formula = '%s ~ %s' % (b, formula)
+        mdf = smf.ols(stats_formula, data=data).fit()
+        fits.append(mdf)
+    return RegressionResults(fits, basis=basis,
+                             feature_names=table.columns)
+
+def gls(formula, table, metadata, tree, **kwargs):
+    """ Generalized Least Squares applied to balances.
+
+    Parameters
+    ----------
+    """
+    pass
+
+def mixedlm(formula, table, metadata, tree, **kwargs):
+    """ Linear Mixed Models applied to balances.
+
+    Parameters
+    ----------
+    """
+    pass
+
+def gee(formula, table, metadata, tree, **kwargs):
+    """ Generalized Estimating Equations applied to balances.
+
+    Parameters
+    ----------
+    """
+    pass
+
--- a/gneiss/_summary.py
+++ b/gneiss/_summary.py
@@ -8,6 +8,7 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
 import pandas as pd
+from skbio.stats.composition import ilr_inv
 
 
 class RegressionResults():
@@ -46,7 +47,10 @@ class RegressionResults():
         self.pvalues = pvals
 
         # calculate the overall R2 value
-
+        sse = sum([r.ssr for r in self.results])
+        ssr = sum([r.ess for r in self.results])
+        sst = sse + ssr
+        self.r2 = 1 - sse / sst
 
     def _check_projection(self, project):
         """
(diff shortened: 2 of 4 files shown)


mortonjt:
Here's the 10,000 ft view of what I'm trying to do

Trying to enable statsmodels statistics to be applied in parallel to all of the balances.  There are at least 5 reasons why I'm creating wrapper objects to do this.

1. There is a non-trivial amount of boiler plate required to format the trees + tables + metadata to make it compatible with `balanceplot`
2. Abstracting out parallelization. The more loops we force users to do, the more prone to error their analyses are (there is a lot of bookkeeping that needs to be done).  Plus, abstracting out this step opens up the possibility of multiprocessing.
3. Enabling projections back into the simplex.  While its great that we can predict the balances between features, we ultimately want to see how these models can predict the original feature probability distributions.  We'd like to be able to transform not only predicted values, but coefficients and residuals.
4. Taking full advantage of the existing framework in statsmodels.  Enabling formula parsing is not trivial.  And statsmodels already does this.
5. There are some statistics that need to be aware of all of the fits on each balances.  (i.e. coefficient of determination aka. R2)

In the end, I would like to be able to run my statistics beginning to end as follows
```python
model = OLS(table, metadata, tree)
results = model.fit()
balanceplot(results['white vs purple'])
```

Now the big question is, how exactly do we do this.

Here's the API that I'm envisioning (influenced by the statsmodels and scikit-learn api).

## _formula.py
- This will contain all of the formulas.  Namely `OLS`, `GLM`, `LME`, `GEE`.  
- All of these modules will have very similar api's.  namely

```
 stats_formula(table : pd.DataFrame, tree : skbio.TreeNode, metadata : pd.DataFrame) -> gneiss.Model
```

- Mainly exists for parsing formulas and processing the user objects.

## _model.py
- This contains information about the underlying statistical model
- There will be two types of models `RegressionModel` and `DifferentialAbundance`
- The `RegressionModel` has the following api

```python
class RegressionModel():
    def __init__(self, models, feature_names, basis)
        """ Creates a regression model given several statsmodels regression models.

        Accepts a list of statsmodels RegressionResults objects
        and performs some addition summary statistics.

        Parameters
        ----------
        models : list, sm objects
            List of statsmodels Regression models.
        feature_names : array_like, str, optional
            List of original names for features.
        basis : np.array, optional
            Orthonormal basis in the Aitchison simplex.
            If this is not specified, then `project` cannot
            be enabled in `coefficients` or `predict`.
        """

    def predict(self, X, project=False, **kwargs):
        """ Performs a prediction based on model.

        Predicts full feature distribution (i.e. distribution of OTUs) given covariates (i.e. ph)

        Parameters
        ----------
        X : pd.DataFrame, optional
            Input table of covariates, where columns are covariates, and
            rows are samples.  If not specified, then the fitted values
            calculated from training the model will be returned.
        project : bool, optional
            Specifies if coefficients should be projected back into
            the Aitchison simplex.  If false, the coefficients will be
            represented as balances  (default: False).
        **kwargs : dict
            Other arguments to be passed into the model prediction.

        Returns
        -------
        pd.DataFrame
            A table of values where rows are coefficients, and the columns
            are either balances or proportions, depending on the value of
            `project`.
        """

    def fit(self, **kwargs):
        """ Fits the regression model.

        Parameters
        ----------
        **kwargs : dict
            Other arguments to be passed into the model prediction.

        Returns
        -------
        pd.DataFrame
            A table of values where rows are coefficients, and the columns
            are either balances or proportions, depending on the value of
            `project`.
        """
```

## _summary.py
This module is primarily focused on summarizing all of the information from the fits.
The api for the `RegressionResults` api is as follows.

```python
    """
    Summary object for storing regression results.
    """
    def __init__(self,
                 stat_results,
                 feature_names=None,
                 basis=None):
        """ Reorganizes statsmodels regression results modules.

        Accepts a list of statsmodels RegressionResults objects
        and performs some addition summary statistics.

        Parameters
        ----------
        stat_results : list, sm.RegressionResults
            List of RegressionResults objects.
        feature_names : array_like, str, optional
            List of original names for features.
        basis : np.array, optional
            Orthonormal basis in the Aitchison simplex.
            If this is not specified, then `project` cannot
            be enabled in `coefficients` or `predict`.
        """
    def coefficients(self, project=False):
        """ Returns coefficients from fit.

        Parameters
        ----------
        project : bool, optional
            Specifies if coefficients should be projected back into
            the Aitchison simplex.  If false, the coefficients will be
            represented as balances  (default: False).

        Returns
        -------
        pd.DataFrame
            A table of values where columns are coefficients, and the index
            is either balances or proportions, depending on the value of
            `project`.

        Raises
        ------
        ValueError:
            Cannot perform projection into Aitchison simplex if `basis`
            is not specified.
        ValueError:
            Cannot perform projection into Aitchison simplex
            if `feature_names` is not specified.
        """
    def residuals(self, project=False):
        """ Returns calculated residuals.

        Parameters
        ----------
        X : pd.DataFrame, optional
            Input table of covariates.  If not specified, then the
            fitted values calculated from training the model will be
            returned.
        project : bool, optional
            Specifies if coefficients should be projected back into
            the Aitchison simplex.  If false, the coefficients will be
            represented as balances  (default: False).
        Returns
        -------
        pd.DataFrame
            A table of values where rows are coefficients, and the columns
            are either balances or proportions, depending on the value of
            `project`.
        """

```

Here are my big questions
1. Does this API make sense?
2. Can we get away with using a single `Model` object and a single `Results` object? Looking at [`GEE`](http://statsmodels.sourceforge.net/devel/generated/statsmodels.genmod.generalized_estimating_equations.GEE.predict.html#statsmodels.genmod.generalized_estimating_equations.GEE.predict) and [`MixedLM`](http://statsmodels.sourceforge.net/devel/generated/statsmodels.regression.mixed_linear_model.MixedLM.html#statsmodels.regression.mixed_linear_model.MixedLM), they are really just regressions, that test for differential abundance.  There are just a few minor difference between these models (i.e. different coefficients), but that's it.
3. What should the file directory look like?  Would a folder called `gneiss/stats`  make sense?

Thanks!!!!

mortonjt:
@antgonza @josenavas @wasade @ElDeveloper any insights about this API will be greatly appreciated!  I'm planning on proceeding with the design that I just listed.  But if there are any glaring problems, its better to resolve them early on.  Thanks!

wasade:
Given your example of what you'd like to be able to do, I don't think the
API supports that? At least I can't identify a mapping between the small
example and the objects/methods

mortonjt:
That's because I had a mistake in my documentation :/

The mapping should be the following
```
formula(table, metadata, tree) -> model
model.fit() -> results
balanceplot(results)
```
Had a mess up in the `fit` module.  That is now corrected.

wasade:
+1 :)

mortonjt:
I've thought about this - let's only focus on regression in this PR.  Differential abundance should be able to be added easily in a following PR.

antgonza:
It sounds/looks fine. Obviously, we will need to review once it's actually implemented. 

Out of curiosity, why _filename vs. just filename?

ElDeveloper:
> 1. Does this API make sense?

It does.

Minor question: the one thing that doesn't make too much sense to me is the naming of `RegressionModel`, seems like it is an orchestrator/summarizer of several models?

> 2. Can we get away with using a single Model object and a single Results object? Looking at GEE and MixedLM, they are really just regressions, that test for differential abundance. There are just a few minor difference between these models (i.e. different coefficients), but that's it.

Probably yes, though it will be valuable to keep an eye open for this, as soon as the feature creep starts showing (if it ever shows), this should be revised. While the intention makes sense, I'm not too familiar with the details and use-cases to actually have a more informative insight.

Just a shot in the dark, but, have you thought about the consequences of using a single object vs multiple in the context of testing, documentation and serialization(not sure if this is needed at all)?

> 3. What should the file directory look like? Would a folder called gneiss/stats make sense?

` ¯\_(ツ)_/¯`, sure.  Given the scope of the project (from what I know), and its current state, I wouldn't worry too much about that level of organization, however the module-level organization seems clean enough. This has the potential of getting complicated, depending on the number of models being maintained.

> I've thought about this - let's only focus on regression in this PR. Differential abundance should be able to be added easily in a following PR.

That sounds like a great idea, fully exploring the implementation space of a single model sounds great, I would actually suggest trying to get the _most_ complicated model and the simplest model that you know of.

mortonjt:
@antgonza I'm adding the `_` to `_model.py` `_summary.py` and `_formula.py`, because I'm thinking making these private methods that are not explicitly available to users.  I was thinking having an import like the following would be intuitive.

```python
from gneiss.stats import gee
```

Still thinking through what file structure will be most intuitive to users/developers.

@ElDeveloper , yes and no.  It is still a single model - but this model is made of up many submodules, one model per balance.  If this is confusing now, we'll definitely want to make this crystal clear in the docs.

Right, which is mainly why I'm outlining the API here.  All of this functionality can be easily performed with the existing API.  However, it requires 100+ lines of boiler plate code, a ton of bookkeeping.  Even I have trouble keeping track of everything.  This is really not sustainable in the long run if others are to use this API.  Realistically, I want to make this as simple to use as possible.

Awesome!  I'll continue pumping out PRs in the next couple of weeks.  In the meanwhile - #28 is really for review.  Thanks!!!

mortonjt:
Removing `_model.py`.  This may be not necessary, since the prediction step can be performed within the `RegressionResults` object.  Note that in statsmodels, this is also where the prediction step takes place.

mortonjt:
Closing in favor of #39

PR #28, opened 2016-07-28
Regression results part2

mortonjt:


mortonjt:
@antgonza @josenavas @wasade review?  Thanks!

gneiss/_summary.py
```diff
@@ -57,3 +58,72 @@ def __init__(self, stat_results,
...
+
+        for r in self.results:
+            c = r.params
+            c.name = r.model.endog_names
+            coef = coef.append(c)
+
+        if project:
+            # `check=True` due to type issue resolved here
+            # https://github.com/biocore/scikit-bio/pull/1396
+            c = ilr_inv(coef.values.T, basis=self.basis, check=False).T
```
antgonza: check=True in the comment but False in the code, which one is it?

gneiss/_summary.py
```diff
@@ -57,3 +58,72 @@ def __init__(self, stat_results,
...
+            c = pd.DataFrame(c, index=self.feature_names,
+                             columns=coef.columns)
+            return c
```
antgonza: what about:
```python
return pd.DataFrame(c, index=self.feature_names, columns=coef.columns)
```
?

gneiss/_summary.py
```diff
@@ -57,3 +58,72 @@ def __init__(self, stat_results,
         # calculate the overall coefficient of determination (i.e. R2)
         sst = sse + ssr
         self.r2 = 1 - sse / sst
+
+    def _check_projection(self, project):
```
antgonza: Are you planning to use this check in other methods? If not, why not add it in the coefficients code (it looks like a really simple check that is happening within `coefficients`)?

antgonza:
some comments

gneiss/_summary.py
mortonjt: This is the reason why I added this comment.

Although this has been resolved in scikit-bio, it is only in the development version - which would mean that this would have to depend on the development version of scikit-bio.

I was thinking about raising an issue, and making the fix at the next release of scikit-bio.  What do you think?

gneiss/_summary.py
mortonjt: 👍

gneiss/_summary.py
mortonjt: The main reason is because a whole bunch of methods are using this check, namely `residuals` and `predict`.  Those are being added in #30

antgonza: OK, thanks. Just out of curiosity, will it make sense to make project a member of the object? What I'm asking is if it will make sense to do somewhere in the code: self.project=True/False, and then just call self.coefficients() & self._check_projection(). This could simplify the code but it will mean that if you are doing self.residuals, self.predict, etc this parameter can't change.

gneiss/_summary.py
antgonza: That fix has been merged in skbio and you could use the dev version of skbio (in travis, setup.py, etc) until there is release, will this work?

mortonjt: Ok. Then this should be ok

gneiss/_summary.py
antgonza: OK, @mortonjt and I discussed this offline and it makes sense to keep as is. Basically, this is converting the results to one of two formats based on the projection.

PR #30, opened 2016-07-30
Regression results part3

mortonjt:
Depends on #28 

Adding in residual and prediction calculations


mortonjt:
Crap, realized that this is pushing into the wrong branch.  Will reopen.

PR #32, opened 2016-07-31
Regression results part3

mortonjt:
Depends on #28


mortonjt:
This is ready for review.  @antgonza @wasade mind if you could take a look?  Thanks!!

gneiss/_summary.py
```diff
@@ -128,3 +129,92 @@ def coefficients(self, project=False):
...
+
+        Parameters
+        ----------
+        X : pd.DataFrame, optional
+            Input table of covariates.  If not specified, then the
+            fitted values calculated from training the model will be
+            returned.
+        project : bool, optional
+            Specifies if coefficients should be projected back into
+            the Aitchison simplex.  If false, the coefficients will be
```
josenavas: Can you add a reference to Aitchison simplex?

sjanssen2:
I don't know how to report this properly, but I am reading the notebook the first time and I want to list some typos, thus I do it here. Please educate me how I would have done this correctly:

1. missing space: species.Below
2. arbituary
3. mistake in the figure? root-node: why is it the ^1/4 and not ^1/3 since the number of leaf nodes is 3
4. can you make the range +1 to -1 for the tree example with three balances?
5. You write about "left" and "right" subtrees, but your example is rotated by 90 degrees, which might confuse the reader.
6. use a consistent writing for "log-ratio" (you have logratios, log ratios and log-ratios)
7. can be run on the_m_ directly?

mortonjt:
Good call.  Let me open up a new pull request for the tutorial, so it'll be easier to comment on changes there.  Thanks!

mortonjt:
Ok.  This is ready for another round of review/merge.  Thanks!

mortonjt:
Can we get another review on this?  Thanks!

gneiss/tests/test_summary.py
```diff
@@ -104,5 +108,100 @@ def test_regression_results_coefficient_project_error(self):
...
+
+        pdt.assert_frame_equal(res_predict, exp_predict)
+
+    def test_regression_results_predict_projection(self):
+        feature_names = ['Z1', 'Z2', 'Z3']
+        basis = _gram_schmidt_basis(3)
+        model = RegressionResults(self.results, basis=basis,
+                                  feature_names=feature_names)
+
+        res_predict = model.predict(self.data[['X']], project=True)
```
wasade: are there any tests for `project=False`? apologize if overlooking something

wasade:
just one comment

PR #33, opened 2016-07-31
OLS

mortonjt:
Depends on #32


gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+    _table, _tree = match_tips(table, tree)
+    non_tips_no_name = [(n.name is None) for n in tree.levelorder()
+                        if not n.is_tip()]
+
+    if any(non_tips_no_name):
+        _tree = rename_internal_nodes(_tree)
+    return _table, _metadata, _tree
+
+
+def _transform(table, tree):
```
wasade: doc string?

gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+def ols(formula, table, metadata, tree, **kwargs):
+    """ Ordinary Least Squares applied to balances.
+
+    A ordinary least square regression is applied to each balance.
+
+    Parameters
+    ----------
+    formula : str
+        Formula representing the statistical equation to be evaluated.
+        These strings are similar to how equations are handled in R.
```
wasade: This sentence only means something to someone who is familiar with how they're handled in R...

wasade: able to point to doc in statsmodels?

gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+# ----------------------------------------------------------------------------
+import pandas as pd
+import statsmodels.formula.api as smf
+from skbio.stats.composition import ilr
+from gneiss.util import match, match_tips, rename_internal_nodes
+from gneiss._summary import RegressionResults
+from gneiss.balances import balance_basis
+
+
+def _process(table, metadata, tree):
```
wasade: would it be better to name this `_intersect_of_table_metadata_tree` or something?

gneiss/tests/test_summary.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+# !/usr/bin/env python
```
wasade: shabang line isnt necessary here and it's incorrect with the space (`# !`)

gneiss/tests/test_formula.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
```
wasade: shabang shouldn't be necessary

wasade:
minor comments, :+1: otherwise

gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+    """ Matches tips, features and samples between the table, metadata
+    and tree.  This module returns the features and samples that are
+    contained in all 3 objects.
+    """
+    _table, _metadata = match(table, metadata)
+    _table, _tree = match_tips(table, tree)
+    non_tips_no_name = [(n.name is None) for n in tree.levelorder()
```
ElDeveloper: Should `tree` here be `_tree`?

gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+from gneiss._summary import RegressionResults
+from gneiss.balances import balance_basis
+
+
+def _process(table, metadata, tree):
+    """ Matches tips, features and samples between the table, metadata
+    and tree.  This module returns the features and samples that are
+    contained in all 3 objects.
+    """
+    _table, _metadata = match(table, metadata)
```
ElDeveloper: Any reason why copies should be created i.e. is there any need to keep table, metadata and tree around?

gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+    _table, _tree = match_tips(table, tree)
```
ElDeveloper: Seems like you would want to match the tree to the already filtered table right? i.e. `_table`

gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+    Returns
+    -------
+    RegressionResults
+        Container object that holds information about the overall fit.
+
+    See Also
+    --------
+    statsmodels.regression.linear_model.OLS
+    """
+    _table, _metadata, _tree = _process(table, metadata, tree)
```
ElDeveloper: Same as above, any reason why you are keeping around copies of the inputs?

gneiss/_formula.py
```diff
@@ -0,0 +1,90 @@
...
+    ilr_table = pd.DataFrame(mat,
+                             columns=non_tips,
+                             index=table.index)
+    return ilr_table, basis
+
+
+def ols(formula, table, metadata, tree, **kwargs):
+    """ Ordinary Least Squares applied to balances.
+
+    A ordinary least square regression is applied to each balance.
```
ElDeveloper: A -> An

gneiss/_formula.py
ElDeveloper: Quick question, would it make sense to raise an exception when either of these three objects turns out to be empty due to a mismatch between for example the metadata and the table?

gneiss/_formula.py
mortonjt: I did reference [patsy](https://patsy.readthedocs.io/en/latest/) in the docstring.  Is that not enough?

gneiss/_formula.py
mortonjt: 👍

gneiss/_formula.py
mortonjt: Good catch.  Thanks!

gneiss/_formula.py
mortonjt: The main reason is because I don't want to have side effects.  It is not desirable to have the original table and metadata modified everytime this function is run.  But this is a good test case to double check.  Thanks!

gneiss/_formula.py
mortonjt: Yeah.  Dumb mistake.  Thanks!

gneiss/_formula.py
mortonjt: Done!

mortonjt:
Thanks for the speedy feedback!  

I had just one question concerning the user interface.  I updated the docstring in `ols` to give a bit more context.

When users typically specify regressions in R, they specify something like as follows

`y ~ x1 + x2 + x3`

where `y` is the response variable, and `x1`, `x2` and `x3` are the explanatory variables (i.e. environmental variables).

Now in this case, we are running a linear regression on each balance.  So if there are `D` species, we run a regression on `D-1` balances.  

For example, if we had 3 balances, the regression equations would look like as follows

`y1 ~ x1 + x2 + x3`
`y2 ~ x1 + x2 + x3`
`y2~ x1 + x2 + x3`

Now, this isn't practical if there are 1000 of balances  - you cannot ask the user to manually input 1000 regression formulas. Which is why that I knocked out the response variable from the equation.

However, it may not be intuitive for the user that a regression is being run if no response variable is being run.  One other idea that I had was maybe to put in a dummy variable like as follows

`balance ~ x1 + x2 + x3`

So that the user will know that the regression is being run directly on the balances.  And have it throw an error if the response variable isn't specified correctly.  

What do you guys think?  Do you think this PR is ok as is, or should we modify the regression formula handling?

wasade:
Meh, see if it actually ends up being confusing going forward and address
if needed?

mortonjt:
👍 Right on.  In that case, all of the comments have been addressed.

gneiss/_formula.py
```diff
@@ -0,0 +1,203 @@
...
+# ----------------------------------------------------------------------------
+import pandas as pd
+import statsmodels.formula.api as smf
+from skbio.stats.composition import ilr
+from gneiss.util import match, match_tips, rename_internal_nodes
+from gneiss._summary import RegressionResults
+from gneiss.balances import balance_basis
+
+
+def _intersect_of_table_metadata_tree(table, metadata, tree):
```
antgonza: Should this have returns, params, etc?

gneiss/_formula.py
antgonza: Perhaps adding a comment about this will make sense to avoid futures questions about it. BTW I think it makes sense to do this and if you really wanted to overwrite it you could:
`table, metadata, tree = _intersect_of_table_metadata_tree(table, metadata, tree)`

gneiss/_formula.py
```diff
@@ -0,0 +1,203 @@
...
+                         'intersection with `metadata`.')
+
+    if any(non_tips_no_name):
+        _tree = rename_internal_nodes(_tree)
+    return _table, _metadata, _tree
+
+
+def _to_balances(table, tree):
+    """ Converts a table of abundances to balances given a tree.
+    Parameters
```
antgonza: new line?

gneiss/_formula.py
```diff
@@ -0,0 +1,203 @@
...
+    table : pd.DataFrame
+        Contingency table where samples correspond to rows and
+        features correspond to columns.
+    tree : skbio.TreeNode
+        Tree object where the leaves correspond to the columns contained in
+        the table.
+
+    Returns
+    -------
+    pd.DataFrame :
```
antgonza: extra `:`?

gneiss/_formula.py
```diff
@@ -0,0 +1,203 @@
...
+        Contingency table where samples correspond to rows and
+        balances correspond to columns.
+    np.array :
```
antgonza: extra `:`?

antgonza:
A few comments.

gneiss/_formula.py
ElDeveloper: Thanks for the explanation, I agree that avoiding side-effects is a good idea. As for avoiding making copies, I was referring to not needing to create `_table` and `_metadata` and instead just referring to `table` by overwriting the variable in return i.e. `table, metadata = match(table, metadata)` instead of `_table, _metadata = match(table, metadata)`. No side-effects and no duplication of variables in the scope of that function.

instead of:

ElDeveloper:
> Meh, see if it actually ends up being confusing going forward and address
if needed?

👍  .... the code looks good.

gneiss/_formula.py
mortonjt: 👍 Done!

mortonjt:
Comments have been addressed!

ElDeveloper:
Thanks @mortonjt.

PR #35, opened 2016-08-09
Correcting typos in ipython notebook

mortonjt:
- [x] missing space: species.Below
- [x] arbituary
- [x] mistake in the figure? root-node: why is it the ^1/4 and not ^1/3 since the number of leaf nodes is 3
- [ ] can you make the range +1 to -1 for the tree example with three balances?
- [x] You write about "left" and "right" subtrees, but your example is rotated by 90 degrees, which might confuse the reader.
- [x] use a consistent writing for "log-ratio" (you have logratios, log ratios and log-ratios)
- [x] can be run on the_m_ directly?


mortonjt:
@sjanssen2 I've addressed 5 of the 7 comments that you suggested.  Could you clarify what you mean in the remaining 2 comments?  Thanks!

sjanssen2:
regarding "can you make the range +1 to -1 for the tree example with three balances?" The generated tree has bars on every internal node, but at one of them the bar grows out of the x-range because it has a negative value, but the x-range is +1 to 0.

regarding "can be run on the directly?" I think that the sentence containing this string is grammatically not correct. I assume that "the" should be "them".

mortonjt:
Right now, I am not sure if it is possible to extend the y-axis labels in ETE.  When I set the parameters for barchart_layout to other than `min_value=0` and `max_value=1`, barchart will not be properly positioned in the tree.

ete_bad


Is there an easy way around this @jhcepas?

jhcepas:
@mortonjt is the issue at StackedBarFace or a rotated BarChartFace?

mortonjt:
The issue is at rotated BarChartFace.

The layout being referenced here is in [`gneiss/layout.py`](https://github.com/biocore/gneiss/blob/master/gneiss/layouts.py#L38).  Thanks!

jhcepas:
oki, it's probably a bug in our side. Will try to fix asap.

mortonjt:
Ok.  Looks like we cannot address the barplot problem until the problem in ete has been fixed.  In the meantime, let's more forward in pushing these changes in.

wasade:
Seems good? Is this PR finalized or are there other components still to add (I see there's one remaining check box...)

mortonjt:
This is ready to go.  The other checkbox cannot be addressed in this PR

jhcepas:
hey, I checked your code and it seems that there is actually no bug to fix on the ETE side. I did not realize that you were using the "float" position. Floating means that the whole BarChartFace (bar and axis labels) will be vertically centered according to the node branch, so the output you get is expected one.  

I guess you tried this already, but you could use a branch-bottom position:
```python 
 t = Tree('((a,b)e, (c,d)f);', format=1)
 b = BarChartFace([0.4], [0], width=5, min_value=-1, max_value=1, label_fsize=4)
 n = t&('f')
 n.add_face(b, 0, position='branch-bottom')
 b.rotation = 270
 t.show()
```
Alternatively, to change the pos of floating objects, you could try modifying its margins (ie. face.margin_left=5).  Pretty much a try/error job

PR #40, opened 2016-08-22
Adding case study 1 notebooks and data

mortonjt:
Showing the balance linear regression on the 88 soils


mortonjt:
This is ready to go!  Would be really nice if we could get these tutorials in for the alpha release.

antgonza:
Do you really need 2 ipython notebooks (https://github.com/biocore/gneiss/pull/41)? I think one with the 2 examples and explaining why they are different will be better, what others think?

mortonjt:
It would be nice to have 2 python notebooks, so that we can demo the linear mixed effects models and the regression separately.  They are completely different methods, so they aren't really comparable.

mortonjt:
We also need both of these notebooks for the paper.

antgonza:
OK, thanks for the explanation. I see 2 possible options:
1. Add a README explaining the difference between the 2 ipython notebooks. Adding an intro to each of them explaining their purpose and having a line or two about the other one.
2. Have a single one with all this info.

What do you think? Happy to talk offline if it will make things faster.

wasade:
@antgonza, how would you feel about a paragraph in the opening cell of each notebook describing what is in the notebook and why? e.g., this notebook shows an example of regression with balances using the 88 soils dataset (link to pub). this notebook reproduces figure X in the manuscript associated with this software repository (manuscript in review), etc etc

antgonza:
I think that's fine + a README in that folder => option 1 of my previous comment. Note that I still think that one single notebook will make things easier: imagine that it walks you thorough all the calculations and plots of the MS.

mortonjt:
@antgonza , I'd prefer to keep these notebooks separate.  Mainly because we are going to be adding in more notebooks in the future, and it'll become more painful to maintain.  Besides, I've always found huge notebooks very annoying to navigate.  

I've added links in the README directly pointing to the IPython notebooks.  Does this address your concerns?

README.md
```diff
@@ -5,10 +5,39 @@
 
 Canonically pronouced *nice*
 
-A compositional statistics and visualization toolbox
 
-You can install Gneiss via pip
+ compositional statistics and visualization toolbox
```
antgonza: I think something got erased by mistake, can you double check?

README.md
```diff
@@ -5,10 +5,39 @@
...
++gneiss is a compositional statistics and visualization toolbox.  
+ 
+Note that gneiss is not compatible with python 2, and is compatible with Python 3.4 or later.
+gneiss is currently in alpha.  We are actively developing it, and __backward-incompatible interface changes can and will arise__.
```
antgonza: Out of curiosity, It sounds like you already know which are these, do you actually know?

mortonjt: At the moment, no.  This is more of a defensive statement.

But I can see this happening.  For instance, we could change the file structure of the repository to clean up the code.

antgonza:
OK, that makes sense and I think is fine. Now, some comments:
- [ ] can you add a link at the top to the README so the user that starts from this notebook knows that there are other notebooks available?
- [ ] can you add a comment about the 100 reads and if this is a parameter that can be used for other analysis or is something that each user should define?
- [ ] I guess convert_biom_to_pandas(table) is something we will have in each notebook and that should be part of biom, right? What are the plans for this?
- [ ] should we use other color, like blue, to highlight important steps, like the 100 reads or sorting the otu and map frames?
- [ ] why 3 more variables? can we add more/less? why do we want to do this?
- [ ] can you add a conclusion section? From the plots it's not super clear that we have accomplished the goal: "Here, we will show that pH by itself can actually predict the microbial abundances, using balance trees."

README.md
antgonza: OK, what about making the statement less strong by removing the "and will arise" ?

antgonza:
Sorry, got it wrong, 1 should have been:
can you add a link at the top to the notebook so the user that starts from this notebook knows that there are other notebooks available?

mortonjt:
Comments addressed. 

The biom issue is ancient and has been raised [here](https://github.com/biocore/biom-format/issues/622).  No idea what the plan is, but we shouldn't let that feature request block this notebook.  I know that there is some nice conversions between biom and pandas available in qiime2 - so maybe we can piggy back on that once we make a qiime2 plugin for gneiss.

I'm not clear on how the conclusion section could be improved - I thought the message was pretty clear.  Do you have suggestions?

mortonjt:
How about adding a previous and next links on the bottom of the notebook?

antgonza:
- previous/next sounds great
- could you add more explanations on how to choose the level (100) and the powers (3)? I'm worry that users will simply follow these numbers
- conclusion: just read the text and it makes more sense now, what about plotting the 2 images side by side and in the same cell and compare them with text?

antgonza:
👍 , can someone else review?

wasade:
On it

README.md
```diff
@@ -5,10 +5,37 @@
...
 ```
 pip install gneiss
 ```
+
+To run through the tutorials, you'll need a few more packages, namely `seaborn`, `biom-format` and `h5py`.
+These packages can be installed with conda as follows
+```
+source activate gneiss
+conda install seaborn h5py
+pip install biom-format
```
wasade: alternatively, `conda install -c bioconda biom-format`

wasade:
In opening, "it was determined that pH is a major driving factor for the microbial communities" -> "it was observed that pH plays a primary role in community assemblage." Most important change is determined -> observed as that was not a mechanistic or causative study. 

The application of the read filter is over both axes such that OTUs with fewer than 100 reads are dropped as well as samples with fewer than 100 reads. The text above omits the filtering over the samples. 

The bolding on "same order" w.r.t. sample metadata might suggest to a reader who is scanning quickly that the sample metadata _are_ in the same order. Suggest: "...**metadata file **are not assured to be in the same order**."

quartic -> quadratic? 

"There cannot exist bacteria that are optimized to live in every possible pH environment." Bold statement. Nature has a habit of throwing wrenches into bold statements.

mortonjt:
Note that quartic is correct, since we are fitting a 4th order polynomial

https://en.wikipedia.org/wiki/Quartic

PR #41, opened 2016-08-22
Adding case study2 notebook and data

mortonjt:
A demo of the linear mixed effects models


mortonjt:
This has already gotten a review from RAQ.  So this should be good to go!

wasade:
Mentioning why there is a `+ 1` on the OTU table data would be useful -- I didn't see a mention of pseudo counts. The last heat map is not well explained as to why its being shown. It seems like its a rhetorical point, and if so it might be worth being more explicit about it. There was also one block of text where there was a double period ".." in the markdown. Probably easier to search for it than to highlight the line in the JSON file

mortonjt:
This is ready for review :)

mortonjt:
This is ready for review.  Note that this is the last PR before uploading this to pypi.

PR #55, opened 2016-10-12
WIP: Parallelization

mortonjt:
Enables parallelization

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # gneiss changelog
 
+## Version 0.1.2
+
+### Features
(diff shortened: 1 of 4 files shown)


mortonjt:
Ok.  This is ready for review.  This enables a nearly linear speed up with the `ols` and `mixedlm` modules.  Testing this on 64 threads, this reduced the runtime for one dataset from nearly 1 hour to a few minutes.

mortonjt:
Also, note that the coveralls message is misleading.  The missed lines are actually being tested in the new parallel unittests

mortonjt:
This is also ready to merge.  The coveralls notification here is actually wrong.  We still have 100% coverage.

mortonjt:
Don't review this yet - there are some problems with joblib for larger jobs

PR #56, opened 2016-10-12
Issue 52

mortonjt:
Adding the `balances` and `tree` attributes to the `RegressionResults` object.

This depends on #55


mortonjt:
This is ready for a quick review.  

The basic premise behind this is to make it easier to analyze the resulting balances.  Before, we required that the balances need to be recalculated for plotting, which could cause problems with consistency (since there are some filters being performed within the module).

This PR makes it more manageable, since all of the balance calculations and trees will be stored within the actual RegressionResults object.

mortonjt:
@antgonza , mind if you could review/merge this PR?  Once this PR is in, we can refactor the all of the analysis notebooks.

mortonjt:
Can we just merge this @wasade @ElDeveloper @antgonza ?  I don't think a full review is required here, since there are really minimal changes being made here.

mortonjt:
Thanks!

Issue #29, opened 2016-07-30
Tree sorting algorithms


mortonjt:
- [ ] Ladderize - sort the subtrees according to how many tips are contained.  Similar to what figtree and Dendroscope do.
- [ ] Gradient sort - similar to what has been implemented in #16 , but on tree leaves rather than tables.

Issue #58, opened 2016-10-16
Optimize tree operations


mortonjt:
- [ ] Optimize `balance_basis` (#8)
- [ ] Optimize `ladderize`
- [ ] Optimize `order_tips`

PR #59, opened 2016-10-16
ENH: Adding two tree sorting methods `ladderize` and `order_tips`

mortonjt:
Addressing #29

### DIFF
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
     table = table.reindex(index=gradient.index,
                           columns=est_niche.index)
     return table
+
+
+def ladderize(tree, ascending=True):
+    """
+    Sorts tree according to the size of the subtrees.
+
+    Parameters
+    ----------
+    tree : skbio.TreeNode
+       Input tree where leafs correspond to features
+       contained in the index in `grouping1.
+    gradient : pd.Series
+       Gradient where the index correspond to features.
+
+    Returns
+    -------
+    skbio.TreeNode
+       A tree whose tips are sorted along the gradient.
+
+    Examples
+    --------
+    >>> from skbio import TreeNode
+    >>> tree = TreeNode.read([u'((a,b)c, ((g,h)e,f)d)r;'])
+    >>> print(tree.ascii_art())
+                        /-a
+              /c-------|
+             |          \-b
+    -r-------|
+             |                    /-g
+             |          /e-------|
+              \d-------|          \-h
+                       |
+                        \-f
+    >>> sorted_tree = ladderize(tree)
+    >>> print(sorted_tree.ascii_art())
+                        /-a
+              /c-------|
+             |          \-b
+    -r-------|
+             |          /-f
+              \d-------|
+                       |          /-g
+                        \e-------|
+                                  \-h
+    """
+    sorted_tree = tree.copy()
+    # Note that this is an extremely slow
+    # See https://github.com/biocore/gneiss/issues/58
+    for n in sorted_tree.postorder(include_self=True):
+        sizes = [len(list(k.tips())) for k in n.children]
+        if ascending:
+            idx = np.argsort(sizes)
+        else:
+            idx = np.argsort(sizes)[::-1]
+        n.children = [n.children[i] for i in idx]
+    return sorted_tree
+
+
+def order_tips(tree, gradient, ascending=True):
+    """
+    Sorts tree according to ordering in tree.
+
+    Parameters
+    ----------
+    tree : skbio.TreeNode
+       Input tree where leafs correspond to features
+       contained in the index in `gradient`.
+    gradient : pd.Series
+       Gradient where the index correspond to feature names.
+       The index in the gradient must be consistent with
+       names of the tips in the `tree`.
+
+    Returns
+    -------
+    skbio.TreeNode
+       A tree whose tips are sorted along the gradient.
+
+    Examples
+    --------
+    >>> from skbio import TreeNode
+    >>> import pandas as pd
+    >>> tree = TreeNode.read([u'((a,b)c, ((g,h)e,f)d)r;'])
+    >>> x = pd.Series({'f':3, 'g':1, 'h':2, 'a':4, 'b':5})
+    >>> print(tree.ascii_art())
+                        /-a
+              /c-------|
+             |          \-b
+    -r-------|
+             |                    /-g
+             |          /e-------|
+              \d-------|          \-h
+                       |
+                        \-f
+    >>> res = order_tips(tree, x)
+    >>> print(res.ascii_art())
+                                  /-g
+                        /e-------|
+              /d-------|          \-h
+             |         |
+    -r-------|          \-f
+             |
+             |          /-a
+              \c-------|
+                        \-b
+    """
+    sorted_tree = tree.copy()
+    # Note that this is an extremely slow
+    # See https://github.com/biocore/gneiss/issues/58
+    for n in sorted_tree.postorder(include_self=True):
+        means = [gradient.loc[list(k.subset())].mean() for k in n.children]
+        if ascending:
+            idx = np.argsort(means)
+        else:
+            idx = np.argsort(means)[::-1]
+        n.children = [n.children[i] for i in idx]
+    return sorted_tree
--- a/gneiss/sort.py
+++ b/gneiss/sort.py
@@ -170,7 +170,7 @@ def ladderize(tree, ascending=True):
 
 def gradient_sort(tree, gradient, ascending=True):
     """
-    Sorts tree according to ordering in tree.
+    Sorts tree according to ordering in gradient.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


mortonjt:
This is ready for review.  Should a quick one.

Note while there is plenty of room to optimize, I have already raised an issue at #58 to for handling larger trees (>10k tips).  This PR is mainly designed for typical studies that have less than 5k tips.
Benchmarks have shown that these operations are very fast for typical studies, but will not scale and likely choke on larger studies.

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+
+def ladderize(tree, ascending=True):
+    """
+    Sorts tree according to the size of the subtrees.
+
+    Parameters
+    ----------
+    tree : skbio.TreeNode
+       Input tree where leafs correspond to features
+       contained in the index in `grouping1.
```
ElDeveloper: What is  `groupin1?

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+    gradient : pd.Series
+       Gradient where the index correspond to features.
```
ElDeveloper: This seems out of date.

mortonjt:
who wants to merge this @wasade @antgonza?  This is pretty thoroughly tested, so this should be quick

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
     table = table.reindex(index=gradient.index,
                           columns=est_niche.index)
     return table
+
+
+def ladderize(tree, ascending=True):
+    """
+    Sorts tree according to the size of the subtrees.
```
ElDeveloper: This description does not match the description of the output.

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+             |          \-b
+    -r-------|
+             |          /-f
+              \d-------|
+                       |          /-g
+                        \e-------|
+                                  \-h
+    """
+    sorted_tree = tree.copy()
+    # Note that this is an extremely slow
```
ElDeveloper: Can you revise this sentence?

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+            idx = np.argsort(sizes)[::-1]
+        n.children = [n.children[i] for i in idx]
+    return sorted_tree
+
+
+def order_tips(tree, gradient, ascending=True):
+    """
+    Sorts tree according to ordering in tree.
+
+    Parameters
```
ElDeveloper: Can you update this section?

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+        sizes = [len(list(k.tips())) for k in n.children]
+        if ascending:
+            idx = np.argsort(sizes)
+        else:
+            idx = np.argsort(sizes)[::-1]
+        n.children = [n.children[i] for i in idx]
+    return sorted_tree
+
+
+def order_tips(tree, gradient, ascending=True):
```
ElDeveloper: This function does pretty much the same thing as `ladderize`, except for the sorting rule it uses. Is it worth factoring that piece out?

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+             |
+             |          /-a
+              \c-------|
+                        \-b
+    """
+    sorted_tree = tree.copy()
+    # Note that this is an extremely slow
+    # See https://github.com/biocore/gneiss/issues/58
+    for n in sorted_tree.postorder(include_self=True):
+        means = [gradient.loc[list(k.subset())].mean() for k in n.children]
```
ElDeveloper: Does it matter if `gradient` and `tree` do not have any matches between the identifiers or if there are more identifiers in one than in the other? i.e. will this lead to any problems?

gneiss/sort.py
antgonza: Is this (`grouping1) accurate? If so, why does't it matter if `grouping1 is not being passed to this func?

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+def ladderize(tree, ascending=True):
+    """
+    Sorts tree according to the size of the subtrees.
+
+    Parameters
+    ----------
+    tree : skbio.TreeNode
+       Input tree where leafs correspond to features
+       contained in the index in `grouping1.
+    gradient : pd.Series
```
antgonza: gradient is not part of this function, right?

gneiss/sort.py
```diff
@@ -114,3 +114,119 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+                       |          /-g
+                        \e-------|
+                                  \-h
+    """
+    sorted_tree = tree.copy()
+    # Note that this is an extremely slow
+    # See https://github.com/biocore/gneiss/issues/58
+    for n in sorted_tree.postorder(include_self=True):
+        sizes = [len(list(k.tips())) for k in n.children]
+        if ascending:
```
ElDeveloper: Minor, but just noticed that this block could be decomposed, by moving `idx = np.argsort(sizes)` outside of the if/else block and then making  somthing like `if not ascending: idx = idx[::-1]`. The same applies for `order_tips`.

gneiss/sort.py
ElDeveloper: Should there be a check that guarantees that `gradient` is of a particular dtype? The error you get when you try to compute the mean of a non-numeric series is rather uninformative.

gneiss/sort.py
mortonjt: 👍

gneiss/sort.py
mortonjt: That is incorrect.  Removing!

gneiss/sort.py
mortonjt: Yup

gneiss/sort.py
mortonjt: Yes!  I agree.  But its not clear to me how to best do this.  I've raised an issue in #60 to further discuss.  But I'd like to get this in first, since it will significantly clean up the notebooks / analyses.

gneiss/sort.py
mortonjt: Good catch.  Adding assertion + test

gneiss/sort.py
mortonjt: How so?  Do you have recommendations on changing the wording?

gneiss/sort.py
```diff
@@ -114,3 +114,118 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+        idx = np.argsort(sizes)
+        if not ascending:
+            idx = idx[::-1]
+        n.children = [n.children[i] for i in idx]
+    return sorted_tree
+
+
+def gradient_sort(tree, gradient, ascending=True):
+    """
+    Sorts tree according to ordering in tree.
```
ElDeveloper: I think you mean Sorts tree according to ordering in gradient.

gneiss/sort.py
ElDeveloper: You are right, this does match the description, 👍!

gneiss/sort.py
```diff
@@ -114,3 +114,118 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+              \d-------|
+                       |          /-g
+                        \e-------|
+                                  \-h
+    """
+    sorted_tree = tree.copy()
+    # Note that this is operation is not optimal
+    # See https://github.com/biocore/gneiss/issues/58
+    for n in sorted_tree.postorder(include_self=True):
+        sizes = [len(list(k.tips())) for k in n.children]
```
wasade: This isn't critical for this PR but this results in a full traversal of a clade for every single node. It would be better to cache ntips across the tree with a single traversal, eg:

```python
def cache_ntips(tree):
    for n in tree.postorder(include_self=True):
        if n.is_tip():
            n_tips = 1
        else:
           n_tips = sum(c.n_tips for c in n.children)
```

gneiss/sort.py
```diff
@@ -114,3 +114,118 @@ def niche_sort(table, gradient, niche_estimator=mean_niche_estimator):
...
+    """
+    sorted_tree = tree.copy()
+    if not np.issubdtype(gradient, np.number):
+        raise ValueError('`gradient` needs to be numeric, not %s' %
+                         gradient.dtype)
+
+    # Note that this operation is not optimal
+    # See https://github.com/biocore/gneiss/issues/58
+    for n in sorted_tree.postorder(include_self=True):
+        means = [gradient.loc[list(k.subset())].mean() for k in n.children]
```
wasade: a similar strategy would be good here for caching the subset, but again not blocking

mortonjt:
Alright.  All comments have been addressed.  Thanks!

wasade:
w00t :) :+1: pending ci

mortonjt:
Its ready!

ElDeveloper:
Thanks @mortonjt.

Issue #61, opened 2016-10-19
Additional regression modules


mortonjt:
- [ ] Generalized estimating equations (gee)
- [ ] General linear model

Issue #52, opened 2016-10-12
Improving the RegressionResults object


mortonjt:
It would be nice if the RegressionResults object just had the tree and balances built inside of it.

There are a few filtering steps done within `ols` and `mixedlm`.  It would be easier to get the resulting tree and balances directly from this object rather than guessing it.

PR #62, opened 2016-10-20
Support for Generalized Estimating Equations

mortonjt:
Addressed part of #61 

This module provides support similar to linear mixed effects models, but is more suited for population studies, rather than subject studies.

Note that this depends on #52 to be merged first, since there is a lot of simplifying functionality in that PR.


mortonjt:
we're closing this in favor for creating a separate q2-plugin

PR #72, opened 2016-11-24
IO for RegressionResults object

mortonjt:
Addressing #49

Issue #18, opened 2016-07-27
Adding sphinx documentation


mortonjt:
- [ ] Add mathjax documentation for `balance_basis`.
- [ ] Make sure that `mean_niche_estimator` renders properly.

PR #73, opened 2016-11-27
Sphinx (#18)

mortonjt:
This addresses #18 and depends on #72 

Note that this is a pretty massive overhaul.  After we get this through and hosted on github pages, this will require a major release.


mortonjt:
Not sure what is going on with coveralls here ...

mortonjt:
The coveralls here is acting weird.  It actually looks like the coverage has increased rather than decreased :/

Regardless.  I don't think that this is a blocker.  Note that we do want to get this out fairly soon, since we want this out before the paper gets published.  Reviews would be greatly appreciated!

mortonjt:
This is good to go.  @RNAer would you like to review/merge this?

RNAer:
👍

PR #84, opened 2016-12-15
Lovell

mortonjt:
Adding in a symmetric version of the lovell proportionality metric.


mortonjt:
Closing this since we won't be using lovell distance

PR #86, opened 2016-12-21
Regression Model refactor part1

mortonjt:
This is just adding some stubs for the new RegressionResults object


mortonjt:
I'm thinking about adding in a `get_balance` with the following signature
```python
get_balance(node):
    """
    Parameters
    ----------
    node : str
         Name of internal node in the tree to be retrieved for

    Returns
    -------
    pd.DataFrame
        Dataframe where the first column contains the numerator and the second column contains
        the deminator of the balance.
    """

@tanaes - was this something along the lines that you were thinking about in #69?
```

tanaes:
I think that would be super useful. But I'm not sure I understand what you mean by the output description. What is the numerator and denominator in this context?

mortonjt:
The numerator and denominator correspond to the actual balance values.

Say if I have a list of balances `[-2, 0, 2]`

The resulting numerator / denominators would be
`0.5*np.array([closure([1, 2]), closure([1, 1]), closure([2, 1])]`
or 
[0.16, 0.25]
[0.25, 0.25] 
[0.25, 0.16]

mortonjt:
Coveralls is just complaining about the abstract method - that's not a problem.  @wasade @tanaes @josenavas would you be willing sanity check this re-architecture?  Thanks!

tanaes:
I'll take a look! I have a hard time wrapping my head around the ETE tree
decorating architecture so should be good practice...

gneiss/_model.py
```diff
@@ -0,0 +1,122 @@
...
+            Orthonormal basis in the Aitchison simplex.
+            Row names correspond to the leafs of the tree
+            and the column names correspond to the internal nodes
+            in the tree. If this is not specified, then `project` cannot
+            be enabled in `coefficients` or `predict`.
+        tree : skbio.TreeNode
+            Bifurcating tree that defines `basis`.
+        balances : pd.DataFrame
+            A table of balances where samples are rows and
+            balances are columns.  These balances were calculated
```
qiyunzhu: Extra space found here.

gneiss/_model.py
```diff
@@ -0,0 +1,122 @@
...
+    def summary(self):
+        """ Print summary results """
+        pass
+
+    def split_balance(self, balance_name):
+        """ Splits a balance into its log ratio components.
+
+        Parameters
+        ----------
+        node : str
```
qiyunzhu: Just curious: the identifier of a node is a string? How does it typically look like?

gneiss/_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+        Returns
+        -------
+        pd.DataFrame
+            Dataframe where the first column contains the numerator and the
+            second column contains the deminator of the balance.
+        """
+        node = self.tree.find(balance_name)
+        left = node.children[0]
+        right = node.children[1]
```
qiyunzhu: The tree is always bifurcating, right?

gneiss/_model.py
```diff
@@ -0,0 +1,122 @@
...
+        b = np.expand_dims(self.balances[balance_name].values, axis=1)
+        # need to scale down by the number of children in subtrees
+        # there is an erroneous factor 1/sqrt(2) from the ilr_inv below
+        # so we'll need to remove that.
+        p = closure(ilr_inv(b) * np.sqrt(2))
+        return pd.DataFrame(p, columns=[left.name, right.name],
+                            index=self.balances.index)
+
+    @classmethod
+    def read_pickle(self, filename):
```
qiyunzhu: This looks a cool solution for me! Is there a reason in the design of the program why both file handle or file path are likely to be parsed to this function?

gneiss/tests/test_model.py
```diff
@@ -0,0 +1,175 @@
...
+import pandas.util.testing as pdt
+import numpy.testing as npt
+
+
+# create some mock classes for testing
+class submock_ok(Model):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def summary(self):
```
qiyunzhu: What's the purpose of this function here?

gneiss/_model.py
mortonjt: Yup!

gneiss/_model.py
mortonjt: Because qiime2 requires all objects that are going to be passed around to be able to be serialized.  I'll be building in a qiime2 plugin into gneiss in an upcoming PR.

gneiss/_model.py
mortonjt: Yup.  It is just the identifier of the internal node.  For instance, if I have an internal node with a name `y1`, that name is also the name of the balance that I am looking for.

gneiss/_model.py
mortonjt: Done

gneiss/tests/test_model.py
mortonjt: The purpose of this function is to force all of the subclasses to implement a summary method.

In the context of regression, we'll want to print out some information about good-of-fit (i.e. r^2),  number of balances, number of coefficients, ...

However, all of this functionality will be implemented in the subclasses - which will come later.

qiyunzhu:
@mortonjt Thanks for answering my questions! All look good.

gneiss/_model.py
```diff
@@ -0,0 +1,119 @@
...
+        filename : str or filehandle
+            Output file to store pickled object.
+        """
+        self._tree = str(self.tree)
+        self.tree = None
+        if isinstance(filename, str):
+            with open(filename, 'wb') as fh:
+                pickle.dump(self, fh)
+        else:
+            pickle.dump(self, filename)
```
wasade: you may want to convert the tree back into newick. for large trees, this probably will bomb due to stack recursion limits

mortonjt: Yup - but I believe that is already addressed here?
I have run into those problems before, which is why I am converting the tree into a string before dumping it into a pickle.

wasade: oh i see now, awesome. recommend deleting `_nodes` though as those references would be stale anyway

mortonjt: oh. I forgot to completely ditch `_nodes`.  We actually don't need that, since the `find` operation in `skbio.TreeNode` will cover that.

gneiss/_model.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        Abstract container for balance models.
+
+        Parameters
+        ----------
+        submodels : list of statsmodels objects
+            List of statsmodels result objects.
+        basis : pd.DataFrame
+            Orthonormal basis in the Aitchison simplex.
+            Row names correspond to the leafs of the tree
```
sjanssen2: typo: leaves

gneiss/_model.py
```diff
@@ -0,0 +1,119 @@
...
+            and the column names correspond to the internal nodes
+            in the tree. If this is not specified, then `project` cannot
```
sjanssen2: what is 'project'?

gneiss/_model.py
sjanssen2: what if balance_name points to a tip and not an internal node?

gneiss/_model.py
mortonjt: The `project` parameter determines if the return type of this method should be proportions or balances.  Literally, it means should I project this into the Simplex (i.e. proportions) or balance space.  

I would be interested in recommendations for improving this aspect of the documentation - this has tripped up a few people in the past.

gneiss/_model.py
mortonjt: Then we would need to throw an error.  Will add that scenario shorly.

mortonjt:
I have addressed the outstanding comments.  @wasade @tanaes @qiyunzhu @sjanssen2 this is ready for another round of review / merge.

gneiss/_model.py
sjanssen2: I'd prefer to add the code for throwing an Error in this PR rather than waiting for future changes. Since it might slip through your to-do list.

mortonjt: Its added :)

See line 65

sjanssen2: Sorry. Just read you comment not the actual code :-/

sjanssen2:
I am fine with this PR!

PR #87, opened 2016-12-23
WIP: Adding dendrogram heatmap

mortonjt:
Adding heatmaps!!!

screen shot 2016-12-22 at 7 14 37 pm

cc @qiyunzhu @tanaes @amnona 

This is still in the brainstorming phase - I'm still not sure how to add tests for this.
Furthermore, I'm not sure how to embed additional dendrogram functionality. It would be nice if the user could directly pass in layouts that the heatmaps could build on top of.


gneiss/plot/_heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+
+   heatmap
+
+"""
+
+import numpy as np
+from ete3 import TreeStyle, AttrFace, ProfileFace
+from ete3 import ClusterNode
+from ete3.treeview.faces import add_face_to_node
+import io
```
qiyunzhu: I guess `import io` should better be before `import numpy`, since `io` is a built-in package of Python.

gneiss/plot/heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+    table.T.to_csv(f, sep='\t', index_label='#Names')
+
+    tr = ClusterNode(str(tree), text_array=str(f.getvalue()))
+    matrix_max = np.max(table.values)
+    matrix_min = np.min(table.values)
+    matrix_avg = matrix_min + ((matrix_max -matrix_min) / 2)
+
+    # Encode the actual profile face
+    nameFace = AttrFace("name", fsize=fsize)
+    def heatmap_layout(node):
```
qiyunzhu: This `heatmap_layout` function has no `return` statement?

gneiss/plot/heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+    ProfileFace.get_color_gradient = get_color_gradient
+    tree.name = ""
+
+    f = io.StringIO()
+    table.T.to_csv(f, sep='\t', index_label='#Names')
+
+    tr = ClusterNode(str(tree), text_array=str(f.getvalue()))
+    matrix_max = np.max(table.values)
+    matrix_min = np.min(table.values)
+    matrix_avg = matrix_min + ((matrix_max -matrix_min) / 2)
```
qiyunzhu: Will there be any concern of `int` vs. `float` conversion here?

gneiss/plot/heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+        params[key] = kwargs.get(key, params[key])
+    fsize = params['rowlabel_size']
+    width = params['width']
+    height = params['height']
+    colorscheme = params['cmap']
+
+
+    # Allow for matplotlib colors to be encoded in ETE3 heatmap
+    # Originally from https://github.com/lthiberiol/virfac
+    def get_color_gradient(self):
```
qiyunzhu: I see that you embedded a function (`get_color_gradient`) in another one (`heatmap`). I didn't know that, but it looks better to me (visually), than placing an independent `get_color_gradient` function before `heatmap`. Is there any advantage of doing so? I will probably do your way too.

gneiss/plot/heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+        from PyQt4 import QtGui
+        try:
+            import matplotlib.pyplot as plt
```
qiyunzhu: My understanding of this is that `matplotlib` is an optional dependency for `gneiss`. Am I right here?

gneiss/plot/_heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+        `ete.tree` as a parameter.
+    cmap: matplotlib colormap
+        String or function encoding matplotlib colormap.
+    labelcolor: str
+        Color of the node labels. (default 'black')
+    rowlabel_size : int
+        Size of row labels. (default 8)
+    width : int
+        Heatmap cell width. (default 200)
+    height : int
```
qiyunzhu: Why heatmap cell height has no default? (should be 14?)

gneiss/plot/heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+    ete.Tree
+        ETE tree object that will be plotted.
+    ete.TreeStyle
+        ETE TreeStyle that decorates the tree and heatmap visualization.
+    """
+
+    params = {'rowlabel_size': 8, 'width':200, 'height':14, 'cmap':'viridis'}
+    for key in params.keys():
+        params[key] = kwargs.get(key, params[key])
+    fsize = params['rowlabel_size']
```
qiyunzhu: Why creating four new variables instead of using `params[x]` directly? (I am just curious. I have no problem with it.)

gneiss/plot/_heatmap.py
mortonjt: oops thanks!

gneiss/plot/heatmap.py
mortonjt: This is because I am anticipating this API to change drastically.  One way ease the process of adding new parameters is to use kwargs that allows for new parameters to be added without breaking the existing API.

In addition, we expect this function to have many parameters in it (perhaps up to 50 parameters) in the near future.  It easier to maintain a dictionary to be passed in rather than 50 parameters in the method header.

gneiss/plot/heatmap.py
mortonjt: Yeah - this is an ETE hack to allow for matplotlib color gradients to be specified in the heatmap.  We could probably move this outside of the heatmap function.  However, I'm currently not seeing a use case for doing so at the moment.

gneiss/plot/heatmap.py
mortonjt: Yup

gneiss/plot/heatmap.py
mortonjt: Not in Python 3 :D 

The `/` operator in python 3 is always floating point division.  If you want integer division in python 3, you use `//`

gneiss/plot/heatmap.py
mortonjt: That is correct.  We are using the layout function to decorate the leaves of the tree.

mortonjt:
@qiyunzhu I have addressed your comments.  Could you merge this if you agree with the edits?

gneiss/plot/heatmap.py
qiyunzhu: Good to know, thanks!

gneiss/plot/heatmap.py
qiyunzhu: I see your rationale.

gneiss/plot/heatmap.py
qiyunzhu: Got it!

PR #88, opened 2017-01-06
Diamondtree

mortonjt:
Adding in plotting functionality that can collapse subtrees and replace the subtrees with diamonds.
In addition, this allows for subtree background coloring.

This functionality can be tested as follows
```python
from gneiss.plot import diamondtree
from skbio import TreeNode
st = TreeNode.read(get_data_path('example.nwk'))
tr, ts = diamondtree(st, collapsers=['y5', 'y18'],
                     breadth_scaling=6, depth_scaling=30,
                     cladecolors={'y5': '#FF0000', 'y18': '#0000FF'},
                     bgcolors={'y29': '#00FF00'})
tr.render(file_name='%%inline', tree_style=ts)
```
To generate the following plot
example_tree

Note that this is a very quick and dirty way to this.  I had trouble getting triangles to work in this context (the triangle orientation wasn't consistent and didn't line up with the tree orientation).  I would like to get triangles and blobs working in the near future.  However due deadlines, it would really nice to get this in first.

@qiyunzhu @tanaes could you take a look at this PR?  @jhcepas if there are any features that you think would be nice to migrate over to ETE3, we can start thinking about how to do so.


gneiss/plot/_blobtree.py
```diff
@@ -0,0 +1,218 @@
...
+        super(_DiamondItem, self).paint(p, option, widget)
+        ete3.treeview.faces._label_painter(self, p, option, widget)
+
+
+class CollapsedDiamondFace(StaticItemFace, Face):
+    """
+    Creates a collapsed node face object.
+
+    """
+    def __init__(self, width, height, label='', color='#0000FF'):
```
wasade: This constructor has a similar API to `_DiamondItem`, would it make sense for them to be the same? The difference is on the default for `label`

gneiss/plot/_blobtree.py
```diff
@@ -0,0 +1,218 @@
...
+              'depth_scaling': 30, 'breadth_scaling': 6,
+              'mode': 'c',
+              # TODO: Enable layout
+              # layout : function, optional
+              #    A layout for formatting the tree visualization. Must take a
+              #    `ete.tree` as a parameter.
+              # For now just define a null function if no layout is defined
+              # TODO: Learning scaling factors for depth and breadth
+              'layout': lambda x: x}
+    for key in params.keys():
```
wasade: no need for `.keys()`

wasade: why not just set all the defaults in the function definition?

gneiss/plot/_blobtree.py
mortonjt: @wasade right.  However, these classes are actually very different because of the classes that they inherit.  
`_DiamondItem` inherits from `QGraphicsPolygonItem` (from PyQT4) and needs to implement `paint` (since it is an abstract method).

`CollapsedDiamondFace` is an ETE object that inherits from both `StaticItemFace` and `Face`.  The main reason why I've have kept these classes separate was because that was the convention in [ETE](https://github.com/etetoolkit/ete/blob/master/ete3/treeview/faces.py).

gneiss/plot/_blobtree.py
mortonjt: Cool.  Removing in the next commit.  The reason why I'm unpacking `**kwargs` in the function definition is because I'm anticipating the parameter space to grow up to 50 parameters.  And this isn't too crazy given what we see in [matplotlib plots](http://matplotlib.org/api/pyplot_api.html).  I'm thinking that adding `**kwargs` would make it easier to iteratively add in new parameters. But I'd be curious to hear your thoughts on this.

gneiss/plot/_blobtree.py
wasade: okay makes sense

gneiss/plot/_blobtree.py
wasade: okay, makes sense

mortonjt:
@wasade I've addressed your comment about the `params.keys()`.  Let me know if you think it is ok!

gneiss/plot/_blobtree.py
```diff
@@ -0,0 +1,218 @@
...
+    ----------
+    .. [1] Hug, Laura A., et al. "A new view of the tree of life."
+       Nature Microbiology 1 (2016): 16048.
+    """
+
+    # TODO: Allow for the option to encode labels in different colors
+    # (i.e. pass in a pandas series)
+    params = {'bgcolors': None, 'cladecolors': '#0000FF',
+              'labelcolor': 'black', 'label_size': 10,
+              'depth_scaling': 30, 'breadth_scaling': 6,
```
tanaes: These depth and breadth scaling parameters don't match the stated defaults.

gneiss/plot/_blobtree.py
```diff
@@ -0,0 +1,218 @@
...
+    life paper [1]. Rather than plotting all of the leaves, specified
+    subtrees will be replaced by diamonds that are scaled to approximate
+    the depth and the width of the subtree.
+
+    Parameters
+    ----------
+    tree : skbio.TreeNode
+        A strictly bifurcating tree defining a hierarchical relationship
+        between all of the features within `table`.
+    collapsers : list of str
```
tanaes: maybe `collapse_nodes` would be a more intuitive name for this?

gneiss/plot/_blobtree.py
mortonjt: Done!

gneiss/plot/_blobtree.py
mortonjt: Done!

mortonjt:
@tanaes I've addressed your comments!  Could you or @wasade merge this PR?

mortonjt:
Thanks @tanaes !

PR #90, opened 2017-01-11
Regression Model refactor part2

mortonjt:
Depends on #86 

This basically defines another subclass `RegressionModel` that will define the structure of the results objects (i.e. `OLS`, `LME`).


gneiss/regression/_model.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(self, *args, **kwargs):
+        """
+        Summary object for storing regression results.
+
+        A `RegressionResults` object stores information about the
+        individual balances used in the regression, the coefficients,
+        residuals. This object can be used to perform predictions.
+        In addition, summary statistics such as the coefficient
+        of determination for the overall fit can be calculated.
+
```
wasade: extra newline

gneiss/tests/test_model.py
```diff
@@ -5,14 +5,13 @@
...
 # ----------------------------------------------------------------------------
-
-
 import pandas as pd
 import statsmodels.formula.api as smf
 from skbio import TreeNode
 from gneiss._model import Model
 import unittest
 import pandas.util.testing as pdt
+import os
```
qiyunzhu: According to [flake8](https://pypi.python.org/pypi/flake8-import-order), I think the order of import should be edited so that standard libraries come first, followed by third-party libraries.

gneiss/regression/_model.py
```diff
@@ -0,0 +1,187 @@
...
+        ----------
+        .. [1] Aitchison, J. "A concise guide to compositional data analysis,
+           CDA work." Girona 24 (2003): 73-81.
+        """
+        coef = pd.DataFrame()
+
+        for r in self.results:
+            c = r.params
+            c.name = r.model.endog_names
+            coef = coef.append(c)
```
qiyunzhu: Looks like there are multiple `append` operations on the data frame. Is there a more efficient way to do it? (Just curious)

gneiss/regression/_model.py
```diff
@@ -0,0 +1,187 @@
...
+        ----------
+        .. [1] Aitchison, J. "A concise guide to compositional data analysis,
+           CDA work." Girona 24 (2003): 73-81.
+        """
+        prediction = pd.DataFrame()
+        for m in self.results:
+            # check if X is none.
+            p = pd.Series(m.predict(X, **kwargs))
+            p.name = m.model.endog_names
+            if X is not None:
```
qiyunzhu: Just curious, what will be the difference between `if X is not None:` and `if X:`?

gneiss/regression/_model.py
```diff
@@ -0,0 +1,187 @@
...
+            # `check=False`, due to a problem with error handling
+            # addressed here https://github.com/biocore/scikit-bio/pull/1396
+            # This will need to be fixed here:
+            # https://github.com/biocore/gneiss/issues/34
+            proj_prediction = ilr_inv(prediction.values.T, basis=self.basis,
+                                      check=False)
+            return pd.DataFrame(proj_prediction,
+                                columns=self.basis.columns,
+                                index=prediction.columns)
+        return prediction.T
```
qiyunzhu: Will it be better if this line is under `else`?

gneiss/regression/_model.py
mortonjt: Done

gneiss/regression/_model.py
mortonjt: I think this use-case should be ok.  This function should not be a heavily used function - so we shouldn't worry about it for the time-being.  If it turns out to be too slow - we can optimize it later.

Right now, I'm getting about 30s for models with 15k features - so I'm not inclined to optimize this for the time being.

gneiss/regression/_model.py
mortonjt: `if X` won't execute the main statement if `X==0` or `X is None`.  `if X is None` only executes if `X is None`

gneiss/regression/_model.py
mortonjt: shouldn't matter - the syntax is the same

gneiss/tests/test_model.py
mortonjt: Done

mortonjt:
Comments addressed.  This is ready for another round of review / merge.  Thanks!

gneiss/regression/_model.py
qiyunzhu: Thanks! `X` is a `pd.DataFrame` so I thought it won't be `0` anyway.

gneiss/regression/_model.py
qiyunzhu: Sounds good!

mortonjt:
Can we get this merged in?  We still have 2 more PRs before we can push out a q2 plugin.  Thanks!

PR #93, opened 2017-01-12
BUGFIX: Allow for no bgcolors and fix get_farthest_leaf issue

mortonjt:
A couple of bug fixes here

1. The bgcolors option now doesn't need to be specified.  Before it was strictly required

2. `get_farthest_leaf` is now being used.  Before `get_farthest_node` was being used, which searched the entire tree, instead of just searching the descendants.


mortonjt:
@tanaes would you mind double checking this before merging?  This should be a very fast review.

mortonjt:
Thanks!

PR #94, opened 2017-01-13
OLS results refactor

mortonjt:
Depends on #90 .  But will need to be merged at the same time as #95 

Also added a regression summary.  Its a little barebones now, but its a good start.


mortonjt:
I'm going to close this PR, since it is very similar to #95

PR #95, opened 2017-01-13
Lme results

mortonjt:
Depends on #94


gneiss/_model.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Abstract container for balance models.
+
+        Parameters
+        ----------
+        submodels : list of statsmodels objects
+            List of statsmodels result objects.
+        basis : pd.DataFrame
+            Orthonormal basis in the Aitchison simplex.
+            Row names correspond to the leafs of the tree
```
tanaes: leaves?

gneiss/_model.py
```diff
@@ -0,0 +1,124 @@
...
+    def summary(self):
+        """ Print summary results """
+        pass
+
+    def split_balance(self, balance_name):
+        """ Splits a balance into its log ratio components.
+
+        Parameters
+        ----------
+        node : str
```
tanaes: maybe 'balance_name'?

gneiss/_model.py
```diff
@@ -0,0 +1,124 @@
...
+            second column contains the deminator of the balance.
+        """
+        node = self.tree.find(balance_name)
+        left = node.children[0]
+        right = node.children[1]
+
+        b = np.expand_dims(self.balances[balance_name].values, axis=1)
+        # need to scale down by the number of children in subtrees
+        # there is an erroneous factor 1/sqrt(2) from the ilr_inv below
+        # so we'll need to remove that.
```
tanaes: Does this need to be addressed with a PR in skbio?

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+from decimal import Decimal
+from statsmodels.iolib.summary2 import Summary
+from collections import OrderedDict
+
+
+def mixedlm(formula, table, metadata, tree, groups, **kwargs):
+    """ Linear Mixed Effects Models applied to balances.
+
+    A linear mixed effects model is performed on nonzero relative abundance
+    data given a list of covariates, or explanatory variables such as ph,
```
tanaes: pH

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+        balances. See `patsy` for more details.
+    table : pd.DataFrame
+        Contingency table where samples correspond to rows and
+        features correspond to columns.
+    metadata: pd.DataFrame
+        Metadata table that contains information about the samples contained
+        in the `table` object.  Samples correspond to rows and covariates
+        correspond to columns.
+    tree : skbio.TreeNode
+        Tree object where the leaves correspond to the columns contained in
```
tanaes: May want to specify explicitly that this is the tree defining the balance space

gneiss/regression/_model.py
```diff
@@ -6,109 +6,40 @@
...
+            using `tree`.
         """
-        if self.basis is None and project:
-            raise ValueError("Cannot perform projection into Aitchison simplex"
-                             " if `basis` is not specified.")
-
-        if self.feature_names is None and project:
-            raise ValueError("Cannot perform projection into Aitchison simplex"
-                             " if `feature_names` is not specified.")
+        super().__init__(*args, **kwargs)
```
tanaes: So LMEModel and RegressionModel both just super() kwargs up to Model? Why have these intermediate levels? Is it just a placeholder for future level-specific code? If so, maybe make a comment to that effect...

gneiss/regression/_ols.py
```diff
@@ -0,0 +1,303 @@
...
+        individual balances. See `patsy` for more details.
+    table : pd.DataFrame
+        Contingency table where samples correspond to rows and
+        features correspond to columns.  The features could either
+        correspond proportions or balances.
+    metadata: pd.DataFrame
+        Metadata table that contains information about the samples contained
+        in the `table` object.  Samples correspond to rows and covariates
+        correspond to columns.
+    tree : skbio.TreeNode
```
tanaes: Similar comment to above -- if this is the balance tree note that explicitly

gneiss/regression/tests/test_mixedlm.py
```diff
@@ -396,7 +244,62 @@ def test_mixedlm_zero_error(self):
...
+        iv = ilr_inv(df[["y1", "y2"]].values)
+        table = pd.DataFrame(iv, columns=['a', 'b', 'c'])
+        metadata = df[['x1', 'x2', 'z1', 'z2', 'v1', 'v2', 'groups']]
+
+        model = mixedlm("x1 + x2", table, metadata, tree, groups="groups",
+                        re_formula="0+z1+z2")
+        model.fit()
+        res = str(model.summary())
+        fname = get_data_path('exp_lme_results.txt')
+        fh = open(fname, 'r')
```
tanaes: Not closing fh -- why not:
```
with open(fname, 'r') as fh:
    exp = fh.read()

self.assertEqual(res, exp)
```

gneiss/regression/tests/test_ols.py
```diff
@@ -0,0 +1,208 @@
...
+        exp_tree = TreeNode.read(['((c,d),(b,a)Y2)Y1;'])
+        metadata = pd.DataFrame({
+            'lame': [1, 1, 1, 1, 1],
+            'real': [1, 2, 3, 4, 5]
+        }, index=['s1', 's2', 's3', 's4', 's5'])
+
+        res = ols('real + lame', table, metadata, tree)
+        res.fit()
+        self.assertEqual(str(table), str(exp_table))
+        self.assertEqual(str(exp_tree), str(tree))
```
tanaes: I don't understand what this is testing. Aren't you explicitly specifying both exp_tree and tree? ...or are you testing that they are not affected by res.fit()?

gneiss/regression/tests/test_ols.py
```diff
@@ -0,0 +1,208 @@
...
+            'real': [1, 2, 3, 4, 5]
+        }, index=['s1', 's2', 's3', 's4', 's5'])
+
+        np.random.seed(0)
+        self.maxDiff = None
+        model = ols('real', table, metadata, tree)
+        model.fit()
+        fname = get_data_path('exp_ols_results.txt')
+
+        fh = open(fname, 'r')
```
tanaes: open with `with` as above?

gneiss/regression/tests/test_ols.py
```diff
@@ -0,0 +1,208 @@
...
+            index=['y0', 'y1'])
+
+        pdt.assert_frame_equal(res_coef, exp_coef,
+                               check_exact=False,
+                               check_less_precise=True)
+        # Double check to make sure the fit is perfect
+        self.assertAlmostEqual(res.r2, 1)
+
+        # Double check to make sure residuals are zero
+        exp_resid = pd.DataFrame([[0., 0.],
```
qiyunzhu: I wonder if there is easier way to create a pandas dataframe with repeated / sequential values? Such as `[0.0] * n` for plain Python lists?

gneiss/regression/tests/test_mixedlm.py
```diff
@@ -396,7 +244,62 @@ def test_mixedlm_zero_error(self):
             'real': [1, 2, 3, 4, 5]
         }, index=['s1', 's2', 's3', 's4', 's5'])
         with self.assertRaises(ValueError):
-            mixedlm('real + lame', table, metadata, tree, groups='lame')
+            res = mixedlm('real + lame', table, metadata, tree, groups='lame')
+            res.fit()
+
+    def test_summary(self):
+        np.random.seed(6241)
```
qiyunzhu: How is this number determined?

gneiss/_model.py
mortonjt: Not really - this is just a result of the ilr transform.  Specifically, balances are calculated as follows

b_i = np.sqrt(L*R / (L+R)) * np.log(g(x) / g(y))

Where L is the number of leaves in the right tree, R is the number of leaves in the left tree, x is the set of proportions in the left tree, y is the set of proportions in the right tree and g represents the geometric mean.  What we want is to dividate out the sqrt term so we extract the geometric means within the log.  I think I know a more straightforward way to do this, so I'll be pushing in a PR into #86 shortly.

gneiss/regression/_model.py
mortonjt: Exactly!  In future iterations of gneiss, we'll want to have support for other types of models such as Classifiers (SVM, Logistic, ...) and Cross correlation (Partial Least Squares, Canonical Correlation Analysis, ...) 

Having this placeholder is critical since there are specific abstract methods that will be common across all model objects (i.e. serialization)

gneiss/regression/tests/test_ols.py
mortonjt: Right, I'm testing that the tree doesn't get changed.  There are some functions within the `ols` that does some trimming, just in case there are samples or otus left out of the OTU table.  These functions will make sure that the metadata and the table have the exact same samples in the same order.  In addition, this will ensure that the otu table and the tree have the exact same OTUs.  

So yeah, we just want to double check to make sure that the trimming isn't doing anything unexpected.

gneiss/regression/tests/test_ols.py
mortonjt: Probably should replace this with np.zeros((5, 2)).  Thanks!

gneiss/regression/tests/test_mixedlm.py
mortonjt: This was borrowed directly from the statsmodels test code.  I'm not sure why this was chosen.

gneiss/regression/_ols.py
mortonjt: Done

mortonjt:
@wasade @sjanssen2 @qiyunzhu @tanaes , this is the last PR that needs to go in before pushing in the qiime2 plugin.  I've merged the OLS and the MixedLM PRs together because they are very similar, and it is just a refactor.  The only major addition is the summary method that summarizes fits and coefficients on each of these objects.  Your feedback here would be greatly appreciated.  Thanks!

qiyunzhu:
Hi @mortonjt I reviewed the code before and it is in good shape to me now. Can you please solve the Travis issue?

qiyunzhu:
Looks good now. @tanaes what do you think?

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+    add a pseudocount.
+
+    Parameters
+    ----------
+    formula : str
+        Formula representing the statistical equation to be evaluated.
+        These strings are similar to how equations are handled in R.
+        Note that the dependent variable in this string should not be
+        specified, since this method will be run on each of the individual
+        balances. See `patsy` for more details.
```
sjanssen2: can you add a link about correct syntax for those strings?

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+        often associated with individuals repeatedly sampled, typically
+        longitudinally.
+    **kwargs : dict
+        Other arguments accepted into
+        `statsmodels.regression.linear_model.MixedLM`
+
+    Returns
+    -------
+    RegressionResults
+        Container object that holds information about the overall fit.
```
sjanssen2: can you give some details about the container?

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+    statsmodels.regression.linear_model.MixedLM
+    skbio.stats.composition.multiplicative_replacement
+    ols
+
+    """
+    table, metadata, tree = _intersect_of_table_metadata_tree(table,
+                                                              metadata,
+                                                              tree)
+    ilr_table, basis = _to_balances(table, tree)
+    data = pd.merge(ilr_table, metadata, left_index=True, right_index=True)
```
sjanssen2: I often have the problem that indices for metadata and the counts have different data type, e.g. int and str. Could you raise an Error if there is no overlap when merging?

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+        of determination for the overall fit can be calculated.
+
+
+        Parameters
+        ----------
+        submodels : list of statsmodels objects
+            List of statsmodels result objects.
+        basis : pd.DataFrame
+            Orthonormal basis in the Aitchison simplex.
+            Row names correspond to the leafs of the tree
```
sjanssen2: leafs -> leaves

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+            and the column names correspond to the internal nodes
+            in the tree. If this is not specified, then `project` cannot
```
sjanssen2: what is 'this' ?

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+            be enabled in `coefficients` or `predict`.
+        tree : skbio.TreeNode
+            Bifurcating tree that defines `basis`.
+        balances : pd.DataFrame
+            A table of balances where samples are rows and
+            balances are columns.  These balances were calculated
```
sjanssen2: whitespace

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+            using `tree`.
+        """
+        super().__init__(*args, **kwargs)
+
+    def fit(self, **kwargs):
+        """ Fit the model
+
+        Parameters
+        ---------
+        regularized : bool
```
sjanssen2: I don't see this parameter explicitly. Is it withing kwargs?

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def summary(self, title=None, yname=None, xname=None, head=None):
+        """ Summarize the Ordinary Least Squares Regression Results.
+
+        Parameters
+        ----------
+        title : string, optional
+            Title for the top table. If not None, then this replaces the
+            default title
+        yname : string, optional
```
sjanssen2: is yname the label for the y axis? If yes, you might want to rename it into ylabel to be more consistent to e.g. matplotlib

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+            Default is `y`
+        xname : list of strings, optional
```
sjanssen2: see above

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+            Default is `var_##` for ## in p the number of regressors
+        head : int
```
sjanssen2: how many can these be in the worst case? Maybe you want to set the default to max. 10 to not accidentally let the user compute tremendous outputs.

gneiss/regression/_mixedlm.py
```diff
@@ -0,0 +1,270 @@
...
+        # Start filling in summary information
+        smry = Summary()
+        # Top results
+        info = OrderedDict()
+        info["No. Observations"] = self.balances.shape[0]
+        info["Model:"] = "Simplicical MixedLM"
+
+        smry.add_dict(info)
+
+        smry.add_title("Simplical Mixed Linear Model Results")
```
sjanssen2: Is simplical a real word?

gneiss/regression/_ols.py
```diff
@@ -0,0 +1,301 @@
...
+                          _to_balances)
+import statsmodels.formula.api as smf
+from statsmodels.iolib.summary2 import Summary
+
+
+# TODO: Register as qiime 2 method
+def ols(formula, table, metadata, tree, **kwargs):
+    """ Ordinary Least Squares applied to balances.
+
+    A ordinary least square regression is performed on nonzero relative
```
sjanssen2: An

gneiss/regression/_ols.py
sjanssen2: looks like a copy from mixedlm, thus look for my comments there

gneiss/regression/tests/test_ols.py
mortonjt: I misread your comment, sorry about that.  It should be `str(res.tree)`, not `str(tree)`.  I'm fixing it now!

gneiss/regression/_mixedlm.py
mortonjt: I did link to the OLS / mixedlm docs in statsmodels.  I can add an additional reference to patsy
https://patsy.readthedocs.io/en/latest/  

You that should cover it?

gneiss/regression/_mixedlm.py
mortonjt: I'm planning on throwing this text together
```
    Returns
    -------
    OLSModel
        Container object that holds information about the overall fit.
        This includes information about coefficients, pvalues, residuals
        and coefficient of determination from the resulting regression.
```

Is this better?

gneiss/regression/_mixedlm.py
mortonjt: That is outdated.  Removing.

gneiss/regression/_mixedlm.py
mortonjt: That was accidentally committed.  Removing.

gneiss/regression/_mixedlm.py
mortonjt: 👍

gneiss/regression/_mixedlm.py
mortonjt: Done - I added the check in `gneiss.util.match`

gneiss/regression/_model.py
mortonjt: I've raised issues #96 and #97 to try to clarify what sorts of methods we're planning on adding in the future.

gneiss/regression/_mixedlm.py
sjanssen2: way better!

mortonjt:
I have carefully addressed all of the outstanding comments.  This is ready for one last round of review.  Thank you for looking over this!  Really appreciate it.

mortonjt:
Thanks!

PR #98, opened 2017-01-19
q2 plugin

mortonjt:
Here's the prototype for the qiime2 plugin.

Note that this also depends on #113 .


gneiss/q2/_type.py
```diff
@@ -0,0 +1,23 @@
...
+Regression = SemanticType('Regression', field_names=['type'])
+
+Linear = SemanticType('Linear',
+                      variant_of=Regression.field['type'])
+LinearMixedEffects = SemanticType('LinearMixedEffects',
+                                  variant_of=Regression.field['type'])
+
+# Tree types
+Hierarchy = SemanticType('Hierarchy', field_names=['type'])
+Cluster = SemanticType('Cluster', variant_of=Hierarchy.field['type'])
```
mortonjt: @jairideout @gregcaporaso this was the original motivation for raising https://github.com/qiime2/q2-types/issues/99

While there are some scenarios that can use a phylogeny for an analysis, most scenarios will deal with different types of hierarchies, whether it be generated from a clustering or something else.

I see a few options for this

1. Define a completely orthogonal type to phylogeny named `Hierarchy`
2. Make `Hierarchy` a super type

Also, would we have to worry about clashing types?  For example, if I have defined `Hierarchy[Cluster]` here, could that potentially clash with an unsupervised sampling clustering type implemented in a classifier package?

jairideout: cc @ebolyen

> Also, would we have to worry about clashing types? For example, if I have defined Hierarchy[Cluster] here, could that potentially clash with an unsupervised sampling clustering type implemented in a classifier package?

Types would only clash if the same name(s) were registered by those two plugins. It might be worth thinking about what names make sense for a classifier plugin so we can come up with distinct names (or reuse the types?).

ebolyen: Just to make sure that this is known, I'd like to point out  that you can define ad-hoc super-types with the `|` operator. e.g. `Phylogony[Rooted] | HierarchicalCluster | Graph`, etc. This might be the best way to go for now, but it doesn't help us with naming the other pieces. 

I like the idea of `Hiearchy`, but what other kinds of variant-types would we have besides `Cluster`? Ideally this could be reused for a classifier, as long as the data was strictly a hierarchy of clusters and not much else.

mortonjt: In context of the balances, there are quite a few more types.

The tree could be produced to represent a gradient along an environment.
It could also be used to represent the largest variance partitions, analogous to Principle Components Analysis.  We have a few more techniques in the works that will yield even more different ways to generate trees specific for features.

For samples, there is hierarchical clustering, but there is also decision trees.  I'd think it would be a good idea to keep trees relating samples and trees relating features to be distinct entities, but maybe stem from the same super type.

gneiss/q2/_type.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+from qiime.plugin import SemanticType
+from q2_types.feature_table import (
+    FeatureTable, Frequency, BIOMV100DirFmt, BIOMV210DirFmt)
+
+
+# Regression types
+Regression = SemanticType('Regression', field_names=['type'])
```
mortonjt: Thinking about changing this to `BalanceRegression` to avoid potential name clashing with other Regression types.  For example there are techniques such as Multinomial Regression or Distance based Regression

mortonjt:
@RNAer have you seen this error before?  It looks like the conda binary cannot be retrieved?

RNAer:
try to restart the travis

mortonjt:
I did - still doesn't work ...

RNAer:
what do you mean? It clearly show you have pep8 error

````
pep8 gneiss setup.py

gneiss/util.py:192:1: E302 expected 2 blank lines, found 1

gneiss/regression/_ols.py:31:1: E302 expected 2 blank lines, found 1

gneiss/regression/tests/test_mixedlm.py:61:1: E302 expected 2 blank lines, found 1

gneiss/regression/tests/test_ols.py:54:5: E303 too many blank lines (2)

make: *** [pep8] Error 1
```

gneiss/regression/_mixedlm.py
```diff
@@ -6,13 +6,24 @@
...
 import statsmodels.formula.api as smf
 from ._model import RegressionModel
-from ._regression import (_intersect_of_table_metadata_tree,
-                          _to_balances)
+from gneiss.util import (_intersect_of_table_metadata_tree,
+                         _to_balances)
 from decimal import Decimal
 from statsmodels.iolib.summary2 import Summary
+from gneiss.util import HAVE_Q2
+if HAVE_Q2:
```
wasade: why not just make gneiss depend on q2?

gneiss/regression/_mixedlm.py
```diff
@@ -260,3 +271,27 @@ def _format(x):
...
+def lme_regression(table: pd.DataFrame, tree: skbio.TreeNode,
+                   metadata: pd.DataFrame, formula: str,
+                   groups: str) -> LMEModel:
+    res = mixedlm(table=table, tree=tree, metadata=metadata,
+                  formula=formula, groups=groups)
+    res.fit()
+    return res
+
+
+plugin.methods.register_function(
```
wasade: ...the downside to not just depending on q2 is that bugs like this can slip through very easily. `plugin` is a `NameError` here.

gneiss/regression/_model.py
```diff
@@ -186,3 +191,25 @@ def pvalues(self):
...
+
+
+# q2 cli
+if HAVE_Q2:
+    Regression_g = SemanticType('Regression_g', field_names=['type'])
+
+    class RegressionFormat_g(model.BinaryFileFormat):
+        def sniff(self):
+            try:
+                res = RegressionModel.read_pickle(open(str(self)))
```
wasade: the sniff is a full parse?

gneiss/util.py
```diff
@@ -184,3 +195,81 @@ def rename_internal_nodes(tree, names=None, inplace=False):
             n.name = label
             i += 1
     return _tree
+
+
+def _intersect_of_table_metadata_tree(table, metadata, tree):
```
wasade: Did this commit sneak in? Doesn't seem like its part of the plugin?

setup.py
```diff
@@ -88,7 +88,13 @@ def finalize_options(self):
           'statsmodels>=0.8.0',
           'ete3',
       ],
+      extra_require={
+          'q2': ['qiime >= 2.0.5', 'biom-format', 'seaborn']
```
wasade: should this be `2017.2.0`?

gneiss/util.py
mortonjt: This is part of refactoring the code.  After evaluating where this sort of functionality really should belong, I think it is more appropriate that it should go into `gneiss.util` since it is a utility function.

mortonjt:
The q2 interface how appears to be working (thanks to @ebolyen, @jakereps and @thermokarst)!

@wasade @sjanssen2 will you be able to review this?  Thanks!!

gneiss/regression/_format.py
```diff
@@ -0,0 +1,31 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+import qiime2.plugin.model as model
+from gneiss.plugin_setup import plugin
+import pickle
+
+
+class RegressionFormat_g(model.BinaryFileFormat):
+    def sniff(self):
+        try:
```
sjanssen2: you also might want to check the type of the pickled object to have a little bit of contents check

gneiss/regression/_regression.py
```diff
@@ -5,86 +5,54 @@
...
-    """
-    non_tips = [n.name for n in tree.levelorder() if not n.is_tip()]
-    basis, _ = balance_basis(tree)
-
-    mat = ilr(table.values, basis=basis)
-    ilr_table = pd.DataFrame(mat,
-                             columns=non_tips,
-                             index=table.index)
-    return ilr_table, basis
+import skbio
```
sjanssen2: isn't that too general? But I don't know how much overhead time it takes to load all of skbio instead of more specific modules.

gneiss/regression/_regression.py
```diff
@@ -5,86 +5,54 @@
...
+
+from q2_composition.plugin_setup import Composition
+from q2_types.feature_table import FeatureTable
+from q2_types.tree import Phylogeny, Rooted, Unrooted
+from qiime2.plugin import Str, Metadata
+from gneiss.plugin_setup import plugin
+from ._type import Regression_g, Linear_g, LinearMixedEffects_g
+
+
+def ols_regression(table: pd.DataFrame, tree: skbio.TreeNode,
```
sjanssen2: cool, a signature like in Haskell :-)

sjanssen2: even if this method is trivial, shouldn't you provide a docstring? Or at least a TODO that the docstring needs to be written in the future?

gneiss/regression/_transformer.py
```diff
@@ -0,0 +1,33 @@
...
+from ._format import RegressionFormat_g
+from gneiss.plugin_setup import plugin
+
+
+@plugin.register_transformer
+def _1(data: OLSModel) -> RegressionFormat_g:
+    ff = RegressionFormat_g()
+    with ff.open() as fh:
+        data.write_pickle(fh)
+    return ff
```
sjanssen2: is it ensured that the filehandle is closed? I recently found that there is a problem in BIOM.

gneiss/regression/_mixedlm.py
```diff
@@ -9,8 +9,8 @@
 import pandas as pd
 import statsmodels.formula.api as smf
 from ._model import RegressionModel
-from ._regression import (_intersect_of_table_metadata_tree,
-                          _to_balances)
+from gneiss.util import (_intersect_of_table_metadata_tree,
+                         _to_balances)
 from decimal import Decimal
 from statsmodels.iolib.summary2 import Summary
```
sjanssen2: line 142 ff are not covered by unittest. Could you please add a test that checks that the ValueError is raised

gneiss/regression/tests/test_format.py
```diff
@@ -0,0 +1,42 @@
...
+
+class TestFormats(TestPluginBase):
+    package = "gneiss.regression.tests"
+
+    def test_regression_format_validate_positive(self):
+        filepath = self.get_data_path('ols.pickle')
+        format = RegressionFormat_g(filepath, mode='r')
+
+        print(filepath, str(format), format.sniff())
+        format.validate()
```
sjanssen2: I still new to unittests in Python. Don't you have to add some kind of self.assertxxx statements to the code?

gneiss/regression/tests/test_regression.py
```diff
@@ -0,0 +1,118 @@
...
+               "--i-tree %s "
+               "--o-linear-model test_ols "
+               "--p-formula 'ph' "
+               "--m-metadata-file %s")
+        proc = subprocess.Popen(cmd % (in_table, in_tree, in_metadata),
+                                shell=True)
+        proc.wait()
+
+        self.assertTrue(os.path.exists("test_ols.qza"))
+
```
sjanssen2: can you also add a test to compare the file contents? Maybe a call to a unix diff?

gneiss/regression/tests/test_regression.py
```diff
@@ -0,0 +1,118 @@
...
+               "--p-formula 'ph' "
+               "--p-groups 'host_subject_id' "
+               "--m-metadata-file %s")
+        proc = subprocess.Popen(cmd % (in_table, in_tree, in_metadata),
+                                shell=True)
+        proc.wait()
+
+        self.assertTrue(os.path.exists("test_lme.qza"))
+        os.remove("test_lme.qza")
+
```
sjanssen2: see above about using diff

gneiss/util.py
```diff
@@ -184,3 +188,81 @@ def rename_internal_nodes(tree, names=None, inplace=False):
...
+    table : pd.DataFrame
+        Contingency table where samples correspond to rows and
+        features correspond to columns.
+    metadata: pd.DataFrame
+        Metadata table that contains information about the samples contained
+        in the `table` object.  Samples correspond to rows and covariates
+        correspond to columns.
+    tree : skbio.TreeNode
+        Tree object where the leaves correspond to the columns contained in
+        the table.
```
sjanssen2: do you mean "the object table" or the metadata-table by "table"?

gneiss/util.py
```diff
@@ -184,3 +188,81 @@ def rename_internal_nodes(tree, names=None, inplace=False):
...
+
+    Returns
+    -------
+    pd.DataFrame
+        Subset of `table` with common row names as `metadata`
+        and common columns as `tree.tips()`
+    pd.DataFrame
+        Subset of `metadata` with common row names as `table`
+    skbio.TreeNode
+        Subtree of `tree` with common tips as `table`
```
sjanssen2: Is it a true sub-tree or a tree where some tips and according branches are removed?

gneiss/regression/_format.py
```diff
@@ -0,0 +1,31 @@
...
+            # just check if the file is pickleable.
+            with open(str(self), 'rb') as fh:
+                pickle.load(fh)
```
wasade: if the pickle is large or complex, this will be an expensive sniff

gneiss/regression/_format.py
```diff
@@ -0,0 +1,31 @@
...
+            return True
+        except:
+            return None
```
wasade: shouldn't this be `False`?

gneiss/regression/_regression.py
wasade: @sjanssen2 it's not really necessary for the qiime2 plugin methods as help text and docs based off the parameters are handled by the plugin API itself based off the info provided at registration

gneiss/regression/_regression.py
```diff
@@ -5,86 +5,54 @@
...
+              formula=formula)
+    res.fit()
+    return res
+
+
+plugin.methods.register_function(
+    function=ols_regression,
+    inputs={'table': FeatureTable[Composition],
+            'tree': Phylogeny[Rooted | Unrooted]},
+    parameters={'formula': Str, 'metadata': Metadata},
```
wasade: Can you add parameter and output descriptions? For example, see [here](https://github.com/qiime2/q2-diversity/blob/master/q2_diversity/plugin_setup.py#L35-L47).

gneiss/regression/_regression.py
```diff
@@ -5,86 +5,54 @@
...
+                   groups: str) -> LMEModel:
+    res = mixedlm(table=table, tree=tree, metadata=metadata._dataframe,
+                  formula=formula, groups=groups)
+    res.fit()
+    return res
+
+
+plugin.methods.register_function(
+    function=lme_regression,
+    inputs={'table': FeatureTable[Composition],
```
wasade: same comment re: descriptions

gneiss/regression/_transformer.py
wasade: Can you open an [issue](https://github.com/biocore/biom-format/issues) with BIOM please? 

In this case, the file handle (`fh`) is closed by the context manager.

gneiss/regression/_transformer.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@plugin.register_transformer
+def _2(ff: RegressionFormat_g) -> OLSModel:
+    with ff.open() as fh:
+        print(fh)
```
wasade: Remove please

gneiss/regression/_transformer.py
```diff
@@ -0,0 +1,33 @@
...
+
+@plugin.register_transformer
+def _3(data: LMEModel) -> RegressionFormat_g:
+    ff = RegressionFormat_g()
+    with ff.open() as fh:
+        data.write_pickle(fh)
+    return ff
+
+
+@plugin.register_transformer
```
wasade: Would it make sense to have a transform from `LMEModel` -> `OLSModel`? It is possible now indirectly via `RegressionFormat_g` but that doesn't mean it's something that should be done...

gneiss/regression/tests/data/not-regression.pickle
```diff
@@ -0,0 +1 @@
+asdfasdfasdf
```
wasade: is this a valid `.pickle` file?

gneiss/regression/tests/test_format.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class TestFormats(TestPluginBase):
+    package = "gneiss.regression.tests"
+
+    def test_regression_format_validate_positive(self):
+        filepath = self.get_data_path('ols.pickle')
+        format = RegressionFormat_g(filepath, mode='r')
+
+        print(filepath, str(format), format.sniff())
```
wasade: Remove please

gneiss/regression/tests/test_format.py
wasade: :+1:

wasade: ...although, I think `validate` just returns `None` if valid and raises if otherwise

gneiss/regression/tests/test_regression.py
```diff
@@ -0,0 +1,118 @@
...
+                                  [0., 0.],
+                                  [0., 0.],
+                                  [0., 0.],
+                                  [0., 0.]],
+                                 index=['s1', 's2', 's3', 's4', 's5'],
+                                 columns=['Y1', 'Y2'])
+        pdt.assert_frame_equal(exp_resid, res.residuals())
+
+    def test_ols_cli(self):
+        # TODO: Is there a better way to test this?
```
wasade: Yes, just load the artifacts and pass them in:

```python
from qiime2.plugins.gneiss.methods import ols_regression

in_table = qiime2.Artifact.load('path/to/foo.qza')
in_tree = qiime2.Artifact.load('path/to/foo.qza')
in_metadata = qiime2.Artifact.load('path/to/foo.qza')

results = ols_regression(table=in_table, tree=in_tree, metadata=in_metadata, etc etc)
```

Note that is a `Results` object. It is a `namedtuple` so you can fetch the individual returned artifacts out easily. If it's just a single artifact coming back, you can implicitly unpack it with:

```python
result, = ols_regression(table=in_table, tree=in_tree, metadata=in_metadata, etc etc)
```

...in which case, `result` would be an artifact of the semantic type as returned by the API method based on the plugin registration.

gneiss/regression/tests/test_regression.py
```diff
@@ -0,0 +1,118 @@
...
+            [[4.211451,  0.0935786, 1.022008, 0.924873],
+             [0.211451,  0.0935786, 1.022008, 0.924873]],
+            columns=['Intercept', 'groups RE', 'x1', 'x2'],
+            index=['Y1', 'Y2'])
+
+        pdt.assert_frame_equal(res.coefficients(), exp_coefficients,
+                               check_less_precise=True)
+
+    def test_lme_cli(self):
+        # TODO: Is there a better way to test this?
```
wasade: see above

gneiss/util.py
```diff
@@ -184,3 +188,81 @@ def rename_internal_nodes(tree, names=None, inplace=False):
             n.name = label
             i += 1
     return _tree
+
+
+def _intersect_of_table_metadata_tree(table, metadata, tree):
+    """ Matches tips, features and samples between the table, metadata
```
wasade: this should be a single sentence for the first line of the doc string

gneiss/util.py
```diff
@@ -184,3 +188,81 @@ def rename_internal_nodes(tree, names=None, inplace=False):
...
+        Subset of `metadata` with common row names as `table`
+    skbio.TreeNode
+        Subtree of `tree` with common tips as `table`
+    """
+    if np.any(table <= 0):
+        raise ValueError('Cannot handle zeros or negative values in `table`. '
+                         'Use pseudocounts or ``multiplicative_replacement``.'
+                         )
+    # check to see if there are overlapping nodes in tree and table
+    overlap = {n.name for n in tree.traverse()} & set(table.columns)
```
wasade: shouldn't this be `tree.tips()`?

mortonjt: Yup - the only problem is we specifically need strings for the names rather than the entire `TreeNode`.  So we'll still need the set comprehension in the end.

gneiss/regression/_transformer.py
mortonjt: I don't think it would make sense to have such a transform, since they are two very distinct objects.  Given this, do you think it is a better idea to create two formats for each of these models?

gneiss/regression/_format.py
mortonjt: Yeah - I couldn't think of a better way to do it, since pickle files don't have headers. 
Do you have a better idea?

gneiss/regression/_format.py
mortonjt: I don't think so

Looking at `NewickFormat`, `get_sniffer` is called.
[get_sniffer](https://github.com/biocore/scikit-bio/blob/master/skbio/io/registry.py#L248) returns `None` if there is no file format detected.

gneiss/regression/_regression.py
mortonjt: I don't think that will be a problem.  `pandas` and `numpy` are much larger projects than `skbio`, but we regularly load all of the modules from those libraries

gneiss/regression/tests/data/not-regression.pickle
mortonjt: Nope - pickle files are binaries.

gneiss/regression/tests/test_regression.py
mortonjt: Thanks!

gneiss/regression/_mixedlm.py
mortonjt: Done

gneiss/regression/tests/test_regression.py
mortonjt: Adding in @wasade 's suggestion

gneiss/regression/_format.py
mortonjt: Also keep in mind that there is an astronomical development burden on reinventing another file format specific for storing Regression objects - so pickle is a strict requirement for the time being.

gneiss/util.py
mortonjt: I don't think I can get this to fit within 1 line.  Do you think it is really necessary for a helper method?

gneiss/regression/_regression.py
mortonjt: Done

gneiss/util.py
mortonjt: the table containing all of the samples.

gneiss/regression/tests/test_regression.py
mortonjt: This has been addressed using the Artifact API.  Really want to avoid using tar here.

gneiss/regression/_format.py
mortonjt: Done

gneiss/regression/_transformer.py
mortonjt: Not sure what is going on - but it seems to appear in quite a few projects

https://github.com/scikit-learn/scikit-learn/issues/3410
https://github.com/python-pillow/Pillow/issues/477
https://github.com/Supervisor/supervisor/issues/391

mortonjt:
The bulk of the comments have been addressed.

Now I have made some large changes here, namely

1. Ditched the `Regression_g` super type and changed `Linear_g` -> `LinearRegression_g`.
2. Split up the `RegressionFormat_g` into two different formats, namely `LinearRegressionFormat_g` and `LinearMixedEffectsFormat_g`.  This ultimately allowed for more fine-tuned checks for the underlying models

This ultimately improves the robustness of these objects and avoids overgeneralizing.

@wasade @sjanssen2 do you mind if you could do one more sanity check once the tests pass? 
We can merge this once you guys think this is ok.  Thanks!

gneiss/regression/tests/test_regression.py
```diff
@@ -0,0 +1,154 @@
...
+        exp_resid = pd.DataFrame([[0., 0.],
+                                  [0., 0.],
+                                  [0., 0.],
+                                  [0., 0.],
+                                  [0., 0.]],
+                                 index=['s1', 's2', 's3', 's4', 's5'],
+                                 columns=['Y1', 'Y2'])
+        pdt.assert_frame_equal(exp_resid, res.residuals())
+
+    def test_ols_cli(self):
```
wasade: I think this is redundant with `test_ols_artifact`?

gneiss/util.py
wasade: What about "The intersection of tips and features" and then the more specific detail follows? I believe samples and metadata subset are implicit as driven by the overlap in tips and features, right?

gneiss/regression/tests/test_regression.py
mortonjt: I think it is ok to keep those tests in, just as a sanity check that the Artifact API / CLI are working as expected.  This is mainly to serve as an alert, especially since the q2 repo is currently in flux.

mortonjt:
🎆  tests passing!

PR #101, opened 2017-01-26
Leave one out cross validation

mortonjt:
There are a few things that are added to this PR namely

- percent_explained : calculates the percentage of the total variance explained for each balance axis (think about the analogue for principal components).

- leave one out cross validation : calculates model error on training dataset and prediction error on testing dataset

- leave one variable out cross validation : calculates R^2 and model error when each variable is left out of the model.  This gives us an idea about the variable importance for each variable.  But more work concerning effect size of each variable will need to be done in the future

In addition, the OLS procedure was slightly optimized to reduce the amount of formula -> matrix conversions that needed to be computed.  Will need to propagate this to the remaining methods.  Thanks @Josh-Lefler for discovering this!

Remember, this is the last major PR that needs to be merged in.  Reviews will be greatly appreciated.  Thanks!

### DIFF
--- a/gneiss/regression/_ols.py
+++ b/gneiss/regression/_ols.py
@@ -21,7 +21,6 @@ import inspect
 
 def _fit_ols(y, x, **kwargs):
     """ Perform the basic ols regression."""
-    #one-time creation of exogenous data matrix allows for faster run-time
     exog_data = x
     submodels = []
     for b in y.columns:
@@ -31,6 +30,7 @@ def _fit_ols(y, x, **kwargs):
         mdf = smf.OLS(endog=endog_data, exog=exog_data, **kwargs)
         submodels.append(mdf)
     return submodels
+
 
 # TODO: Register as qiime 2 method
 def ols(formula, table, metadata, tree, **kwargs):
@@ -172,6 +172,7 @@ def ols(formula, table, metadata, tree, **kwargs):
                                                               tree)
     ilr_table, basis = _to_balances(table, tree)
     ilr_table, metadata = ilr_table.align(metadata, join='inner', axis=0)
+    # one-time creation of exogenous data matrix allows for faster run-time
     x = dmatrix(formula, metadata, return_type='dataframe')
     submodels = _fit_ols(ilr_table, x)
     return OLSModel(submodels, basis=basis,
@@ -359,7 +360,7 @@ class OLSModel(RegressionModel):
             params[key] = kwargs.get(key, params[key])
             kwargs.pop(key, None)
 
-        nobs = self.balances.shape[0] # number of observations (i.e. samples)
+        nobs = self.balances.shape[0]  # number of observations (i.e. samples)
         cv_iter = LeaveOneOut(nobs)
         endog = self.balances
         exog_names = self.results[0].model.exog_names
@@ -422,7 +423,7 @@ class OLSModel(RegressionModel):
 
             kwargs.pop(key, None)
 
-        nobs = self.balances.shape[0] # number of observations (i.e. samples)
+        nobs = self.balances.shape[0]  # number of observations (i.e. samples)
 
         endog = self.balances
         exog_names = self.results[0].model.exog_names
@@ -442,10 +443,9 @@ class OLSModel(RegressionModel):
                 res_i = [r.fit(**kwargs) for r in res_i]
             # See `statsmodels.regression.linear_model.RegressionResults`
             # for more explanation on `ess` and `ssr`.
-            # sum of squares regression. Also referred to as
-            # explained sum of squares.
+            # sum of squares regression.
             ssr = sum([r.ess for r in res_i])
-            # sum of squares error.  Also referred to as sum of squares residuals
+            # sum of squares error.
             sse = sum([r.ssr for r in res_i])
             # calculate the overall coefficient of determination (i.e. R2)
             sst = sse + ssr
@@ -453,7 +453,7 @@ class OLSModel(RegressionModel):
             # degrees of freedom for residuals
             dfe = res_i[0].df_resid
             results.loc[feature_id, 'mse'] = sse / dfe
-            i+=1
+            i += 1
         return results
 
     def percent_explained(self):
@@ -462,4 +462,3 @@ class OLSModel(RegressionModel):
         # instead of population variance (df=0).
         axis_vars = np.var(self.balances, ddof=1, axis=0)
         return axis_vars / axis_vars.sum()
-
--- a/gneiss/regression/_ols.py
+++ b/gneiss/regression/_ols.py
@@ -352,8 +352,8 @@ class OLSModel(RegressionModel):
 
         for i, (inidx, outidx) in enumerate(cv_iter):
             sample_id = self.balances.index[i]
-            res_i = _fit_ols(y=endog.loc[inidx], x=exog.loc[inidx], **kwargs)
-            res_i = [r.fit(**kwargs) for r in res_i]
+            model_i = _fit_ols(y=endog.loc[inidx], x=exog.loc[inidx], **kwargs)
+            res_i = [r.fit(**kwargs) for r in model_i]
 
             # mean sum of squares error
             sse = sum([r.ssr for r in res_i])
--- a/gneiss/regression/_ols.py
+++ b/gneiss/regression/_ols.py
@@ -16,7 +16,6 @@ import statsmodels.formula.api as smf
 from statsmodels.iolib.summary2 import Summary
 from statsmodels.sandbox.tools.cross_val import LeaveOneOut
 from patsy import dmatrix
-import inspect
 
 
 def _fit_ols(y, x, **kwargs):
@@ -422,8 +421,6 @@ class OLSModel(RegressionModel):
             params[key] = kwargs.get(key, params[key])
 
             kwargs.pop(key, None)
-
-        nobs = self.balances.shape[0]  # number of observations (i.e. samples)
 
         endog = self.balances
         exog_names = self.results[0].model.exog_names
--- a/gneiss/regression/_ols.py
+++ b/gneiss/regression/_ols.py
@@ -20,7 +20,6 @@ from patsy import dmatrix
 
 def _fit_ols(y, x, **kwargs):
     """ Perform the basic ols regression."""
-    exog_data = x
     # mixed effects code is obtained here:
     # http://stackoverflow.com/a/22439820/1167475
     submodels = [smf.OLS(endog=y[b], exog=x, **kwargs) for b in y.columns]
--- a/gneiss/regression/_ols.py
+++ b/gneiss/regression/_ols.py
@@ -7,14 +7,30 @@
 # ----------------------------------------------------------------------------
 from decimal import Decimal
 from collections import OrderedDict
-
+import numpy as np
 import pandas as pd
 from gneiss.regression._model import RegressionModel
 from ._regression import (_intersect_of_table_metadata_tree,
                           _to_balances)
 import statsmodels.formula.api as smf
 from statsmodels.iolib.summary2 import Summary
+from statsmodels.sandbox.tools.cross_val import LeaveOneOut
+from patsy import dmatrix
+import inspect
 
+
+def _fit_ols(y, x, **kwargs):
+    """ Perform the basic ols regression."""
+    #one-time creation of exogenous data matrix allows for faster run-time
+    exog_data = x
+    submodels = []
+    for b in y.columns:
+        endog_data = y[b]
+        # mixed effects code is obtained here:
+        # http://stackoverflow.com/a/22439820/1167475
+        mdf = smf.OLS(endog=endog_data, exog=exog_data, **kwargs)
+        submodels.append(mdf)
+    return submodels
 
 # TODO: Register as qiime 2 method
 def ols(formula, table, metadata, tree, **kwargs):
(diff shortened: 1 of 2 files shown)


gneiss/regression/_ols.py
```diff
@@ -7,13 +7,28 @@
...
 import statsmodels.formula.api as smf
 from statsmodels.iolib.summary2 import Summary
+from statsmodels.sandbox.tools.cross_val import LeaveOneOut
+from patsy import dmatrix
+
+
+def _fit_ols(y, x, **kwargs):
+    """ Perform the basic ols regression."""
+    exog_data = x
+    submodels = []
```
josenavas: Minor, and it depends on how big `len(y.columns)` is but this can be written as:
```python
submodels = [smf.OLS(endog=y[b], exog=x, **kwargs) for b in y.columns]
```

Appending to a list may take more than expected if the list is really long.

gneiss/regression/_ols.py
```diff
@@ -298,3 +318,144 @@ def r2(self):
 
         sst = sse + ssr
         return 1 - sse / sst
+
+    @property
+    def mse(self):
```
josenavas: Is this property described somewhere (i.e. doc)?

gneiss/regression/_ols.py
```diff
@@ -298,3 +318,144 @@ def r2(self):
...
+           pred_err : np.array, float
+               Prediction mean sum of squares error for each iteration of
+               the cross validation.
+
+        See Also
+        --------
+        fit
+        """
+        params = {'fit_regularized': False}
+        for key in params:
```
josenavas: Are you expecting to grow `params` in the future? As it is, this `for` loop only executes one operation so I don't see the point of having it.

gneiss/regression/_ols.py
```diff
@@ -298,3 +318,144 @@ def r2(self):
...
+        nobs = self.balances.shape[0]  # number of observations (i.e. samples)
+        cv_iter = LeaveOneOut(nobs)
+        endog = self.balances
+        exog_names = self.results[0].model.exog_names
+        exog = pd.DataFrame(self.results[0].model.exog,
+                            index=self.balances.index,
+                            columns=exog_names)
+        results = pd.DataFrame(index=self.balances.index,
+                               columns=['mse', 'pred_err'])
+        i = 0
```
josenavas: I think there are 2 more pythonic way of doing this is, I'm unsure which one is the best for your case. Since `i` is just an index, this should work:
```python
for sample_id, (inidx, outidx) in zip(self.balances.index, cv_iter):
   etc...
```

However, If you still need `i` for something else that I may be missing, then this is the pythonic way of doing the index counting:

```python
for i, (inidx, outidx) in enumerate(cv_iter):
    ....
```
And remove the `i+=1` from the end of the for loop.

gneiss/regression/_ols.py
```diff
@@ -298,3 +318,144 @@ def r2(self):
...
+    def lovo(self, **kwargs):
+        """ Leave one variable out cross-validation.
+
+        Calculates summary statistics for each iteraction of leave one variable
+        out cross-validation, specially `r2` and `mse` on entire model.
+        This technique is particularly useful for feature selection.
+
+        Parameters
+        ----------
+        regularized : bool
```
josenavas: `regularized` is not named parameter of this function

gneiss/regression/_ols.py
```diff
@@ -298,3 +318,144 @@ def r2(self):
...
+        pd.DataFrame
+           Rsquared : np.array, flot
+               Coefficient of determination for each variable left out.
+           mse : np.array, float
+               Mean sum of squares error for each iteration of
+               the cross validation.
+        """
+
+        params = {'regularized': False}
+        for key in params:
```
josenavas: Same comment as above regarding the for loop.

gneiss/regression/_ols.py
```diff
@@ -298,3 +318,144 @@ def r2(self):
...
+        endog = self.balances
+        exog_names = self.results[0].model.exog_names
+        exog = pd.DataFrame(self.results[0].model.exog,
+                            index=self.balances.index,
+                            columns=exog_names)
+        cv_iter = LeaveOneOut(len(exog_names))
+        results = pd.DataFrame(index=exog_names,
+                               columns=['mse', 'Rsquared'])
+        i = 0
+        for inidx, outidx in cv_iter:
```
josenavas: Same comment as above regarding the usage of the `i` variable.

gneiss/regression/tests/test_ols.py
```diff
@@ -226,6 +240,113 @@ def test_summary_head(self):
...
+        exp_coefs = pd.DataFrame({
+            'y0': [-2.074816e+09, -2.039271e+08, -1.843143e+08,
+                   2.261170e+09, -7.925699e+06],
+            'y1': [-2.074816e+10, -2.039271e+09, -1.843143e+09,
+                   2.261170e+10, -7.925699e+07]
+            }, index=['Intercept', 'x1', 'x2', 'x3', 'x4']).T
+        # The L1 regularization either has numerical issues
+        # or is random ...
+        res_coefs = res.coefficients()
+        pdt.assert_index_equal(exp_coefs.index, res_coefs.index)
```
josenavas: Minor and I'm unsure if it applies in the pandas testing functions: in general we use `assert*(observed, expected)`, but I'm unsure if the pandas error messages are expecting this convention.

gneiss/regression/tests/test_ols.py
```diff
@@ -226,6 +240,113 @@ def test_summary_head(self):
...
+        res = ols(formula="x1 + x2 + x3 + x4",
+                  table=self.y, metadata=self.x, tree=self.t2)
+        res.fit(regularized=True)
+        exp_coefs = pd.DataFrame({
+            'y0': [-2.074816e+09, -2.039271e+08, -1.843143e+08,
+                   2.261170e+09, -7.925699e+06],
+            'y1': [-2.074816e+10, -2.039271e+09, -1.843143e+09,
+                   2.261170e+10, -7.925699e+07]
+            }, index=['Intercept', 'x1', 'x2', 'x3', 'x4']).T
+        # The L1 regularization either has numerical issues
```
josenavas: From my ignorant view, if it is random, can you set a seed? Otherwise, can you test some of the values up to a certain decimal position?

gneiss/regression/_ols.py
mortonjt: I didn't know that.  Thanks!

gneiss/regression/_ols.py
mortonjt: I'll add it in.  It is a measure of Mean of Sum of Squares Error, which is used for calculating R^2 and the cross validation statistics.

gneiss/regression/_ols.py
mortonjt: The main reason for having the kwargs argument was to pass these into the `fit` function, which have a ton of parameters in them.  One advantage of doing this is if the `fit` function in statsmodels changes, we aren't forced to change our APIs to fit their changes.

gneiss/regression/_ols.py
mortonjt: Will use enumerate instead.  Thanks!

gneiss/regression/_ols.py
mortonjt: I'll move this up to the params section.  We do actually need this

gneiss/regression/tests/test_ols.py
mortonjt: So this is actually depending on https://github.com/statsmodels/statsmodels/issues/3393

Not sure what is going on there - but maybe we shouldn't include any regularization fitting here until that issue gets resolved.  I'm going to remove those fits in the next commit.

gneiss/regression/tests/test_ols.py
mortonjt: Going to remove this test since we won't be needing it.

gneiss/regression/_ols.py
mortonjt: Scratch that.  I'll delete it.

gneiss/regression/_ols.py
```diff
@@ -203,11 +210,24 @@ def __init__(self, *args, **kwargs):
...
+            when performing the fit. (default = False)
+        **kwargs : dict
+           Keyword arguments used to tune the parameter estimation.
+
+        """
         for s in self.submodels:
             # assumes that the underlying submodels have implemented `fit`.
-            m = s.fit(**kwargs)
+            if regularized:
+                m = s.fit(**kwargs)
```
mortonjt: This is backwards.  This should be `fit_regularized`,  not `fit`.  I'm removing the `fit_regularized` functionality for the time being to resolve this.  But this actually messed up the `test_mse` test.  I have fixed that in the upcoming commit

josenavas:
Once tests pass

mortonjt:
@sjanssen2 @qiyunzhu could you review / merge?  Thanks!

gneiss/regression/_ols.py
```diff
@@ -7,13 +7,23 @@
...
 from statsmodels.iolib.summary2 import Summary
+from statsmodels.sandbox.tools.cross_val import LeaveOneOut
+from patsy import dmatrix
+
+
+def _fit_ols(y, x, **kwargs):
+    """ Perform the basic ols regression."""
+    # mixed effects code is obtained here:
+    # http://stackoverflow.com/a/22439820/1167475
+    submodels = [smf.OLS(endog=y[b], exog=x, **kwargs) for b in y.columns]
```
qiyunzhu: Maybe these two lines can be merged into one? (Not sure if what I proposed is Pythonic, if not sorry.)

gneiss/regression/_ols.py
```diff
@@ -203,12 +205,20 @@ def __init__(self, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
 
-    def fit(self, **kwargs):
-        """ Fit the model """
-        for s in self.submodels:
-            # assumes that the underlying submodels have implemented `fit`.
-            m = s.fit(**kwargs)
-            self.results.append(m)
+    def fit(self, regularized=False, **kwargs):
```
qiyunzhu: So this rewritting is basically an optimization of the old function. Is that right?

gneiss/regression/_ols.py
```diff
@@ -295,6 +303,122 @@ def r2(self):
...
+        exog_names = self.results[0].model.exog_names
+        exog = pd.DataFrame(self.results[0].model.exog,
+                            index=self.balances.index,
+                            columns=exog_names)
+        cv_iter = LeaveOneOut(len(exog_names))
+        results = pd.DataFrame(index=exog_names,
+                               columns=['mse', 'Rsquared'])
+        for i, (inidx, outidx) in enumerate(cv_iter):
+            feature_id = exog_names[i]
+            res_i = _fit_ols(endog, exog.loc[:, inidx], **kwargs)
```
qiyunzhu: This part is bit awkward to me. `res_i` is defined twice with different data types. Is this Pythonic?

gneiss/regression/_ols.py
```diff
@@ -295,6 +303,122 @@ def r2(self):
...
+            # sum of squares regression.
+            ssr = sum([r.ess for r in res_i])
+            # sum of squares error.
+            sse = sum([r.ssr for r in res_i])
+            # calculate the overall coefficient of determination (i.e. R2)
+            sst = sse + ssr
+            results.loc[feature_id, 'Rsquared'] = 1 - sse / sst
+            # degrees of freedom for residuals
+            dfe = res_i[0].df_resid
+            results.loc[feature_id, 'mse'] = sse / dfe
```
qiyunzhu: Can this `dfe` be zero?

gneiss/regression/_ols.py
```diff
@@ -295,6 +303,122 @@ def r2(self):
...
+        return results
+
+    def percent_explained(self):
+        """ Proportion explained by each principal balance."""
+        # Using sum of squares error calculation (df=1)
+        # instead of population variance (df=0).
+        axis_vars = np.var(self.balances, ddof=1, axis=0)
+        return axis_vars / axis_vars.sum()
```
qiyunzhu: Can this `axis_vars.sum()` be zero?

gneiss/regression/_ols.py
mortonjt: If that ever happens, would indicate overfitting to the extreme (i.e. more parameters than data points).  I think it is safe to say that this will never happen.

gneiss/regression/_ols.py
mortonjt: That would mean that the variance is zero.  For practical reasons, that will not happen.

gneiss/regression/_ols.py
mortonjt: yup!

gneiss/regression/_ols.py
mortonjt: meh sure.

gneiss/regression/_ols.py
mortonjt: ok, I'll rename it

mortonjt: Basically, all this is doing is

1. Retrieving the model specifications for the linear regression (`model_i`)
2. Actually fitting the model with a numerical optimization (`res_i`)

qiyunzhu: Sounds good!

gneiss/regression/_ols.py
qiyunzhu: Got it!

gneiss/regression/_ols.py
qiyunzhu: I see.

Issue #109, opened 2017-02-03
Port pycogent drawing 


mortonjt:
Looks like pycogent has support for coloring trees
https://github.com/pycogent/pycogent/blob/master/cogent/draw/dendrogram.py

Combine this with bokeh, and we can have interactive trees
http://chuckpr.github.io/blog/trees2.html

rob-knight:
That's a great idea! The cogent layout algorithms are pretty good and should include wedges etc, may need to backport some functionality from topiaryexplorer for algorithms for dynamic collapsing.

On Feb 2, 2017, at 6:24 PM, Jamie Morton <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks like pycogent has support for coloring trees
https://github.com/pycogent/pycogent/blob/master/cogent/draw/dendrogram.py

Combine this with bokeh, and we can have interactive trees
http://chuckpr.github.io/blog/trees2.html

-

PR #110, opened 2017-02-08
Dendrogram objects

mortonjt:
Addressing #109 

These objects will allow for the creation of different tree layouts and the retrieval of the node plotting coordinates.  Ultimately, this will allow for any plotting program to consume these coordinates.

Note that this is a direct port from pycogent.

- [x] Unittests

EDIT: 

Bokeh works with the coordinates!!! 

screen shot 2017-02-07 at 3 59 48 pm

I'll be incorporating another PR to follow up with the bokeh support shortly.

Otherwise. this PR is now complete.  @tanaes @amnona would you like to take a look at this PR?


gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+# Copyright (c) 2016--, gneiss development team.
+#
+# Distributed under the terms of the GPLv3 License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+"""Drawing trees.
+
+Draws horizontal trees where the vertical spacing between taxa is
```
sjanssen2: is this code about the construction of the tree representation or a layouting algorithm?
More suitable wording might be: Draws trees horizontally, i.e. root left, tips right, ....

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+constant.  Since dy is fixed dendrograms can be either:
```
sjanssen2: what is dy?

sjanssen2: why do you speak about dendrograms now? I thought you call those objects trees?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+# and constant-y is suitable for placing alongside a sequence alignment anyway.
+from skbio import TreeNode
+import pandas as pd
+import numpy
+import abc
+
+
+class Dendrogram(TreeNode):
+    # One of these for each tree edge.  Extra attributes:
+    #    depth - distance from root to bottom of edge
```
sjanssen2: what is the bottom of an edge? The vertical coordinate of the lower end of an edge?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+        Parameters
+        ----------
+        use_lengths: bool
+            Specifies if the branch lengths should be included in the
+            resulting visualization (default True).
+        """
+        super().__init__(**kwargs)
+        self.use_lengths_default = use_lengths
+
+    def _cache_ntips(self):
```
sjanssen2: why not using TreeNode.count(tips=True)

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+        width : int
+            The width of the canvas.
+
+        Returns
+        -------
+        pd.DataFrame
+            index : str
+                name of node
+            x : float
+                x-coordinate of point
```
sjanssen2: point --> node

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+            y : float
+                y-coordinate of point
+            left : str
```
sjanssen2: is it necessary to restrict this code to bifurcating trees?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+                name of left child node
+            right : str
+                name of right child node
+        """
+        self.rescale(width, height)
+        result = {}
+        for node in self.postorder(include_self=True):
```
sjanssen2: isn't include_self by default set to True?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+                name of node
+            x : float
+                x-coordinate of point
+            y : float
+                y-coordinate of point
+            left : str
+                name of left child node
+            right : str
+                name of right child node
+        """
```
sjanssen2: what about is_tip?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+            is_tip = {'is_tip': node.is_tip()}
+            result[node.name] = {**coords, **children, **is_tip}
+        result = pd.DataFrame(result).T
+
+        # reorder so that x and y are first
+        cols = list(result)
+        cols.insert(0, cols.pop(cols.index('y')))
+        cols.insert(0, cols.pop(cols.index('x')))
+        result = result.ix[:, cols]
+        return result
```
sjanssen2: I haven't tested it, but what about this one liner?
return results.loc[:,['x','y'] + list(set(result.columns) - set(['x','y']))]

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+                best_scale = scale
+                mid_x = width / 2 - ((max(xs) + min(xs)) / 2) * scale
+                mid_y = height / 2 - ((max(ys) + min(ys)) / 2) * scale
+                best_args = (scale, mid_x, mid_y, direction, angle)
+
+        self.update_coordinates(*best_args)
+        return best_scale
+
+    def update_coordinates(self, s, x1, y1, a, da):
+        """ Update x, y coordinates of tree nodes in canvas.
```
sjanssen2: Can you give an example for which event it is necessary to update the coordinates?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+        ----------
+        s : float
+            scaling
+        x1 : float
+            x midpoint
+        y1 : float
+            y midpoint
+        a : float
+            direction (degrees)
+        da : float
```
sjanssen2: what is the difference between angle and direction?

gneiss/plot/_dendrogram.py
qiyunzhu: I guess that dx and dy are the coordinates of a node in the tree.

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+                y-coordinate of point
+            left : str
+                name of left child node
+            right : str
+                name of right child node
+        """
+        self.rescale(width, height)
+        result = {}
+        for node in self.postorder(include_self=True):
+            children = {'child%d' % i: n.name
```
qiyunzhu: What's the point of adding prefix "child" to each node? Is it to differ from other categories of nodes (e.g., "parent")?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+                        for i, n in enumerate(node.children)}
+            coords = {'x': node.x2, 'y': node.y2}
+            is_tip = {'is_tip': node.is_tip()}
+            result[node.name] = {**coords, **children, **is_tip}
+        result = pd.DataFrame(result).T
+
+        # reorder so that x and y are first
+        cols = list(result)
+        cols.insert(0, cols.pop(cols.index('y')))
```
qiyunzhu: I guess Python' `reversed` function can make the code bit simpler.

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+        best_scaling : float
+            largest scaling factor in which the tree can fit in the canvas.
+
+        Notes
+        -----
+        This function has a little bit of recursion.  This will
+        need to be refactored to remove the recursion.
+        """
+
+        angle = 2*numpy.pi / self._n_tips
```
qiyunzhu: The latest pep8 plugin of Atom will complain if not `2 * numpy.pi`.

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+        # this loop is a horrible brute force hack
+        # there are better (but complex) ways to find
+        # the best rotation of the tree to fit the display.
+        best_scale = 0
+        for i in range(60):
+            direction = i / 60.0 * numpy.pi
+            points = self.update_coordinates(1.0, 0, 0, direction, angle)
+            xs = [x for (x, y) in points]
```
qiyunzhu: How about `xs, ys = zip(*points)`?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+            ys = [y for (x, y) in points]
+
+            # double check that the tree fits within the margins
+            scale = min(float(width) / (max(xs) - min(xs)),
+                        float(height) / (max(ys) - min(ys)))
+            scale *= 0.95  # extra margin for labels
```
qiyunzhu: How was the factor 0.95 decided? Will the labels always occupy an absolute amount of space (e.g., 12 px) or relative (e.g., 1 em)?

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+        (self.x1, self.y1, self.x2, self.y2, self.angle) = (x1, y1, x2, y2, a)
+        # TODO: Add functionality that allows for collapsing of nodes
+        a = a - self._n_tips * da / 2
+        if self.is_tip():
+            points = [(x2, y2)]
+        else:
+            points = []
+            # recurse down the tree
+            for child in self.children:
+                ca = child._n_tips * da
```
qiyunzhu: What is `ca`? (add comment?)

gneiss/plot/_dendrogram.py
```diff
@@ -0,0 +1,235 @@
...
+constant.  Since dy is fixed dendrograms can be either:
+ - square:     dx = distance
+ - not square: dx = max(0, sqrt(distance**2 - dy**2))
+
+Also draws basic unrooted trees.
+
+For drawing trees use either:
+ - UnrootedDendrogram
+
+Note: This is directly ported from pycogent.
```
wasade: Please note the version, the file, and explicitly state the license pycogent

gneiss/plot/_dendrogram.py
wasade: Doing that per-node would be very expensive as it would require full traversals of each subtree. This post-order aggregation requires only a single traversal of the tree

gneiss/plot/_dendrogram.py
mortonjt: I believe that pycogent is GPL.  Do you think that is a problem?

gneiss/plot/_dendrogram.py
mortonjt: Its both?  The main point of this to provide a layout algorithm.  However, in order to do so, we need a tree representation.  To handle that, we are subclass the skbio TreeNode object, and just assigning attributes directly to nodes in the tree.

gneiss/plot/_dendrogram.py
mortonjt: dx and dy refer to the spacing between the points within the tree layout.  This documentation was also directly ported pycogent.  Probably should ditch that since it is dated ...

mortonjt: Dendrograms in this context are just subclassed trees that contain the layout representation.

gneiss/plot/_dendrogram.py
mortonjt: I don't think so?  The `y` coordinate may not be at the end of the dataframe.  We just want to reorder it so that the dataframe looks nicer.

gneiss/plot/_dendrogram.py
mortonjt: That will probably work.  Thanks!

gneiss/plot/_dendrogram.py
mortonjt: That is just to specify the name of the column.  For instance, if I had the following tree

```
          /-a
         |
-r-------|--b
         |
          \-c
```
The returned dataframe would be

```
   x  y child0 child1 child2
r  (x)  (y)       a      b      c
```
where a, b, c are children, and x and y represent the coordinates to be plotted in the plot.
We just want some pointer information to keep track of who's who's children, so that we can plot all of the edges in the tree.

gneiss/plot/_dendrogram.py
mortonjt: yeah ...

gneiss/plot/_dendrogram.py
mortonjt: Nope.  That documentation is wrong.  It will be fixed.

gneiss/plot/_dendrogram.py
mortonjt: ¯\_(ツ)_/¯ Those were the margins specified in pycogent. I'll add a TODO to investigate this later.

gneiss/plot/_dendrogram.py
mortonjt: Two scenarios

1. This is called within the `rescale` function, where the best orientation / scaling is chosen to fit the tree in the proper dimensions
2. Dynamic rendering of wedges, so that we can represent some of the subtrees as wedges

gneiss/plot/_dendrogram.py
mortonjt: a = angle
da = difference in angle

gneiss/plot/_dendrogram.py
mortonjt: I'm assuming that this is the arc encapsulating the subtree.

gneiss/plot/_dendrogram.py
mortonjt: In other words, da is the angle resolution of the tree

gneiss/plot/_dendrogram.py
mortonjt: Done

gneiss/plot/_dendrogram.py
mortonjt: Done.  And knocked out all cases similar to this.

gneiss/plot/_dendrogram.py
mortonjt: Done

gneiss/plot/_dendrogram.py
mortonjt: Done

gneiss/plot/_dendrogram.py
mortonjt: 👍

gneiss/plot/_dendrogram.py
mortonjt: Not sure.  I removed that documentation, since it appears to be outdated and doesn't reflect the current state of this refactor

qiyunzhu:
Thanks for the explanations! @mortonjt

mortonjt:
Ok.  This is ready for another round of review.  Note that I just found out that statsmodels did a major release a few days ago, which is why I have changed the statsmodels requirement in `setup.py`

mortonjt:
Thanks!

PR #112, opened 2017-02-11
WIP: Radialplot

mortonjt:
Depends on #110 

This creates unrooted radial tree visualizations using bokeh.  

Right now, this creates an interactive tree, with zoom, reset and name highlighting capabilities.

This is attempting to follow the seaborn paradigm, using the tree and the internal data structure for plotting instead of a pandas dataframe.

In the future, we'll need to create methods that'll make it easy to perform bulk operations to append values to scikit-bio trees.  Maybe some methods that take in `pd.DataFrame` and decorates all of the nodes and edges in the tree with those attributes.  That is a future PR.

There are a couple of concerns that will need to be addressed here.
- [x] Unittests.  Not sure how to do so with bokeh just yet.
- [ ] The interactive component could provide a substantial amount of customizability, but the current setup limits that, due to the explosion of parameters required to enable that.   Another option is to not include the interactive component, and let the user configure it themselves.


gneiss/plot/_radial.py
```diff
@@ -0,0 +1,144 @@
...
+    """ Plots unrooted radial tree.
+
+    Parameters
+    ----------
+    tree : instance of skbio.TreeNode
+       Input tree for plotting.
+    node_hue : str
+       Name of variable in `tree` to color nodes.
+    node_size : str
+       Name of variable in `tree` that species the radius of nodes.
```
qiyunzhu: Typo: `species` => `specifies`

gneiss/plot/_radial.py
```diff
@@ -0,0 +1,144 @@
...
+               node_alpha='node_alpha', edge_hue='edge_hue',
+               edge_alpha='edge_alpha', edge_width='edge_width',
+               figsize=(500, 500), **kwargs):
+    """ Plots unrooted radial tree.
+
+    Parameters
+    ----------
+    tree : instance of skbio.TreeNode
+       Input tree for plotting.
+    node_hue : str
```
qiyunzhu: I think the usage of term "hue" here isn't right. You actually passed the entire RGB value to `hue`. But to be strict hue is a single number that should be used with saturation and brightness.

gneiss/plot/_radial.py
```diff
@@ -0,0 +1,144 @@
...
+    bifurcate
+    """
+    # This entire function was motivated by
+    # http://chuckpr.github.io/blog/trees2.html
+    t = UnrootedDendrogram.from_tree(tree)
+
+    nodes = t.coords(figsize[0], figsize[1])
+
+    # fill in all of the node attributes
+    default_node_hue = '#D3D3D3'
```
qiyunzhu: Maybe consider putting a comment here what color (in English) does '#D3D3D3' mean?

gneiss/plot/_radial.py
```diff
@@ -0,0 +1,144 @@
...
+    nodes[node_hue] = pd.Series({n.name: getattr(n, node_hue,
+                                                 default_node_hue)
+                                 for n in t.levelorder(include_self=True)})
+
+    default_node_size = 1
```
qiyunzhu: Is there a way to zip these two iterations into one? (I don't know I am just asking)

gneiss/plot/_dendrogram.py
```diff
@@ -41,9 +46,34 @@ def __init__(self, use_lengths=True, **kwargs):
...
+        else:
+            self.length = self.length
+
+        self.depth = (depth or 0) + self.length
+
+        children = self.children
+        if children:
+            for c in children:
+                c.update_geometry(use_lengths, self.depth)
+            self.height = max([c.height for c in children]) + self.length
```
mortonjt: `height` is the total height of the subtree.  `length` is the branch length connecting to that subtree.

mortonjt: I'll update the docstring for it and specify it in the attributes

gneiss/plot/_radial.py
```diff
@@ -0,0 +1,144 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+import pandas as pd
+from gneiss.plot._dendrogram import UnrootedDendrogram
+
+try:
+    from bokeh.models.glyphs import Circle, Segment
+    from bokeh.models import ColumnDataSource, DataRange1d, Plot
+    from bokeh.models import HoverTool, BoxZoomTool, ResetTool
+except ImportError:
```
mortonjt: I'm trying to avoid my dependencies blowing up - so all of the visualization code that I'm using will be based on optional dependencies.

qiyunzhu: Sounds good. I also wonder how to set up Travis to test both scenarios. Can one install / not install a package simultaneously in .travis.yaml?

mortonjt:
I think it is ok to only install bokeh on travis.

gneiss/plot/_radial.py
mortonjt: What do you think about something like as follows
```
    def _retreive(tree, x, default):
        return pd.Series({n.name: getattr(n, x, default)
                          for n in tree.levelorder()})

    nodes[node_hue] = _retreive(t, node_hue, '#D3D3D3')

```

gneiss/plot/_radial.py
mortonjt: done

gneiss/plot/_radial.py
mortonjt: I replaced `hue` with `color`.  What are your thoughts?

mortonjt:
@qiyunzhu all of the comments have been addressed.

Would you like to clone it and try this out yourself?  There's the following commands that you can run to test within a notebook.

```python
# import stuff
from gneiss.plot import radialplot

import numpy as np
from scipy.cluster.hierarchy import ward
from skbio import TreeNode, DistanceMatrix
from bokeh.io import output_notebook, show
from gneiss.plot._dendrogram import UnrootedDendrogram
import bokeh.plotting as plt
output_notebook()

# create random tree
np.random.seed(0)
num_otus = 3  #otus 
x = np.random.rand(num_otus)
dm = DistanceMatrix.from_iterable(x, lambda x, y: np.abs(x-y))
lm = ward(dm.condensed_form())
t = TreeNode.from_linkage_matrix(lm, np.arange(len(x)).astype(np.str))
t = UnrootedDendrogram.from_tree(t)

# populate tree with colors and stuff
for i, n in enumerate(t.postorder(include_self=True)):
    if not n.is_tip():
        n.name = "y%d" % i
        n.color='#00FF00'
        n.edge_color='#FF0000'
        n.node_size=10
    else:
        n.color='#0000FF'
        n.edge_color='#00FF00'
        n.node_size=10
    n.length = np.random.rand()*3
    n.edge_width = 2

# plot and show
p = radialplot(t, node_hue='color', edge_hue='edge_color', 
               node_size='node_size', edge_width='edge_width')
show(p)
```

qiyunzhu:
I tested the code and it worked. The plot looks cool. @mortonjt Ready to ship.

PR #138, opened 2017-03-07
Adding environment.yml for binder

mortonjt:
This does the following

1. Adding environment.yml for binder
2. Fixing warnings
3. Deprecating diamondtree in favor of radialplot


mortonjt:
@antgonza would you like to merge this?  This is required to get binder up and running.

mortonjt:
We'll also push out another minor release after this - still testing out the new set of dependencies.

gneiss/balances.py
```diff
@@ -34,9 +34,8 @@
 try:
     from ete3 import Tree, TreeStyle
     from gneiss.layouts import default_layout
-except ImportError:
-    raise ImportWarning('ete3 is not installed.  '
-                        'ETE3 style visualizations will not be available.')
+except:
```
antgonza: why this change? perhaps printing the error will be useful for users

gneiss/layouts.py
```diff
@@ -7,9 +7,8 @@
 # ----------------------------------------------------------------------------
 try:
     from ete3 import faces, AttrFace, CircleFace, BarChartFace
-except ImportError:
-    raise ImportWarning('ete3 is not installed.  '
-                        'ETE3 style visualizations will not be available.')
+except:
```
antgonza: same: why and printing error useful for user?

gneiss/plot/_blobtree.py
```diff
@@ -5,225 +5,9 @@
...
-            return x[node.name], True
-        elif x is None:
-            return "", False
-        else:
-            raise TypeError("color type %s not supported." % type(x).__name__)
-    except KeyError:
-        # Use default if the color isn't specified
-        c = default_color
-        return c, False
+    warnings.warn("`diamondtree` has been deprecated.  "
```
antgonza: if this is beta, is it really worth to keep this or simply remove?

ipynb/cfstudy/cfstudy.ipynb
```diff
@@ -11,11 +11,79 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 4,
    "metadata": {
     "collapsed": false
    },
-   "outputs": [],
+   "outputs": [
```
antgonza: do you really want to add all these tracebacks to your notebook?

gneiss/balances.py
mortonjt: This will still print an error if the function is called. 

Apparently `raise ImportWarning` raises an exception - which breaks the entire CLI interface.
Another options is to use `warnings.warn(...)`.  However, this isn't what we want, since this warning will be printed every time a command line call is made.

mortonjt: Here's an example of what this error looks like

screen shot 2017-03-07 at 7 06 10 am

Do you think this is acceptable?

gneiss/plot/_blobtree.py
mortonjt: Let's remove it.

gneiss/balances.py
antgonza: yup, thanks for clarifying.

mortonjt:
@antgonza thank you for reviewing  this one - that could have been bad if we merged in all of the tracebacks.  And I think the import errors are a little more clear now.

I think it should ok once the tests pass.  Thanks!

PR #150, opened 2017-04-09
Ilr transform

mortonjt:
The main purpose of this PR is to expose the ilr functionality so that other plugins can take balances directly as input.  

With these balances, standard alpha/beta diversity + regression + other machine learning techniques should be able to applied directly.

There was some refactoring done in addition.


mortonjt:
@RNAer do you mind if you could review / merge this PR?  This is only exposing the `ilr` function from skbio to the cli.

RNAer:
sure. will take a look later today

mortonjt:
Thanks!

PR #153, opened 2017-04-10
Testing utils

mortonjt:
This is a really simple PR to simulate band tables (as shown in the horseshoe paper) and block diagonal matrices.


gneiss/util.py
```diff
@@ -317,3 +317,74 @@ def random_tree(n):
...
+
+        mat[lower_row:upper_row, lower_col:upper_col] = B
+
+    # Make last block fill in the remainder
+    B = np.random.uniform(size=(nrows-upper_row, ncols-upper_col))
+    mat[upper_row:, upper_col:] = B
+    return mat
+
+
+def _shift(l, n):
```
wasade: can an english description of this algorithm be provided? is it just rolling the vector? if so, can `np.roll` be used?

gneiss/tests/test_util.py
```diff
@@ -345,6 +347,37 @@ def test_random_tree(self):
...
+    def test_block_diagonal_4x4(self):
+        np.random.seed(0)
+        res = block_diagonal(4, 4, 2)
+        exp = np.array([[0.5488135, 0.71518937, 0., 0.],
+                        [0.60276338, 0.54488318, 0., 0.],
+                        [0., 0., 0.4236548, 0.64589411],
+                        [0., 0., 0.43758721, 0.891773]])
+        npt.assert_allclose(res, exp, rtol=1e-5, atol=1e-5)
+
+    def test_block_diagonal_3x4(self):
```
wasade: what happens if a `nblocks == 1`?

gneiss/util.py
mortonjt: Fixing

gneiss/tests/test_util.py
mortonjt: It should throw an error, because that is outside the scope of this function

mortonjt:
@wasade comments addressed.  Thanks for reviewing!

mortonjt:
@wasade can you double check this?

PR #156, opened 2017-04-16
Balance diagnostic plots

mortonjt:
Adding some diagnostic plots via seaborn to make the partitions a little more intuitive.

Here's the typical use-case for the boxplot
```python
import pandas as pd
from skbio import TreeNode
df = pd.DataFrame({
            'y': [-2, -2.2, -1.8, -1.5, -1, 1, 1.5, 2, 2.2, 1.8],
            'group': ['a', 'a', 'a', 'a', 'a', 'b', 'b', 'b', 'b', 'b'],
            'hue': ['0', '1', '0', '1', '0', '1', '0', '1', '0', '1']}
        )
tree = TreeNode.read(['((c, d)z, (b,a)x)y;'])
a = balance_boxplot(balance_name='y', y='group', hue='hue', data=df)
```

image

And here's the use case for the barplots

```python
feature_df = pd.DataFrame(
            {
                'type': ['tomato', 'carrots', 'apple', 'bacon'],
                'food': ['vegatable', 'vegatable', 'fruit', 'meat'],
                'seed': ['yes', 'no', 'yes', 'no']
            },
            index=["a", "b", "c", "d"]
        )
ax_denom, ax_num = balance_barplots(tree, 'y', header='food',
                                            feature_metadata=feature_df)
```

image


gneiss/plot/_decompose.py
```diff
@@ -0,0 +1,137 @@
...
+    seaborn.boxplot
+    """
+    import seaborn as sns
+    if ax is None:
+        f, ax = plt.subplots()
+
+    # the number 20 is pretty arbitrary - we are just
+    # resizing to make sure that there is separation between the
+    # edges of the plot, and the boxplot
+    pad = (data[balance_name].max() - data[balance_name].min()) / 20
```
tanaes: Does Seaborn / mpl really not have good implicit axis range determination?

mortonjt: It actually does - but what really sucks here is when you try to do the background shading.

So you do need to have some padding in order to make sure that there is no white space at the ends.   Of course, I would be open to a more clean solution.

mortonjt:
@tanaes I'm going go ahead and merge all of the PRs that have been already approved - its becoming really messy to maintain with all open PRs.

PR #151, opened 2017-04-10
Heatmap cleanup

mortonjt:
The two main things that are done in this PR

1. Making line width an exposed parameter
2. Making matplotlib colormaps an exposed parameter

Here's some examples of how to run the qiime commands and their resulting outputs in view.qiime2.org
```
qiime gneiss dendrogram-heatmap \
    --i-table 88soils_composition.biom.qza \
    --i-tree ph_tree.nwk.qza \
    --m-metadata-file 88soils_rounded_metadata.txt \
    --m-metadata-category "ph_rounded" \
    --o-visualization "viridis_heatmap"
```
screen shot 2017-04-09 at 6 03 47 pm

```
qiime gneiss dendrogram-heatmap \
    --i-table 88soils_composition.biom.qza \
    --i-tree ph_tree.nwk.qza \
    --m-metadata-file 88soils_rounded_metadata.txt \
    --m-metadata-category ph_rounded \
    --p-color-map "seismic" \
    --o-visualization "seismic_heatmap"
```
screen shot 2017-04-09 at 6 03 51 pm

```
qiime gneiss dendrogram-heatmap \
    --i-table 88soils_composition.biom.qza \
    --i-tree ph_tree.nwk.qza \
    --m-metadata-file 88soils_rounded_metadata.txt \
    --m-metadata-category ph_rounded \
    --p-color-map "seismic" \
    --p-method "log" \
    --o-visualization "log_seismic_heatmap"
```
screen shot 2017-04-09 at 6 03 58 pm

And here's the help menu
```
tests-MacBook-Pro-4:gneiss mortonjt$ qiime gneiss dendrogram-heatmap --help
Usage: qiime gneiss dendrogram-heatmap [OPTIONS]

  Visualize the feature tables as a heatmap. with samples sorted along a
  specified metadata category and features clustered together specified by
  the tree.In addition, the heatmap values are clr transformed.

Options:
  --i-table PATH                  Artifact: FeatureTable[Composition]
                                  [required]
                                  The feature table that will be
                                  plotted as a heatmap. This table is assumed
                                  to have strictly positive values.
  --i-tree PATH                   Artifact: Phylogeny[Rooted]  [required]
                                  A
                                  hierarchy of feature identifiers where each
                                  tipcorresponds to the feature identifiers in
                                  the table. This tree can contain tip ids
                                  that are not present in the table, but all
                                  feature ids in the table must be present in
                                  this tree.
  --m-metadata-file PATH          Metadata mapping file  [required]
[...]

PR #157, opened 2017-04-18
Dendrogram cleanup

### LINKED ISSUE #128: Rectangular coordinates for tree
There are some scenarios where representing tree balances as triangles will cause overlaps.

screen shot 2017-03-04 at 8 56 17 pm

mortonjt:
This depends on #151 so merge that first.

This PR resolves #128 .  This can be tested with the following notebook

```python
from gneiss.plot import heatmap
from gneiss.plot._heatmap import _sort_table

import pandas as pd
import pandas.util.testing as pdt
from skbio import TreeNode, DistanceMatrix
from scipy.cluster.hierarchy import ward
from gneiss.plot._dendrogram import SquareDendrogram
import numpy as np
%matplotlib inline

np.random.seed(0)
table = pd.DataFrame(np.random.random((5, 5)),
                     index=['0', '1', '2', '3', '4'],
                     columns=['0', '1', '2', '3', '4'])

num_otus = 5  # otus
x = np.random.rand(num_otus)
dm = DistanceMatrix.from_iterable(x, lambda x, y: np.abs(x-y))
lm = ward(dm.condensed_form())
t = TreeNode.from_linkage_matrix(lm, np.arange(len(x)).astype(np.str))
t = SquareDendrogram.from_tree(t)
md = pd.Series(['a', 'a', 'a', 'b', 'b'],
               index=['0', '1', '2', '3', '4'])
for i, n in enumerate(t.postorder()):
    if not n.is_tip():
        n.name = "y%d" % i
    n.length = np.random.rand()*3

highlights = pd.DataFrame({'y8': ['#FF0000', '#00FF00'],
                           'y6': ['#0000FF', '#F0000F']}).T

fig = heatmap(table, t, md, highlights)
```

image


gneiss/plot/_heatmap.py
```diff
@@ -73,10 +73,21 @@ def heatmap(table, tree, mdvar, highlights=None, cmap='viridis',
     edges = edges.unstack()
     edges = pd.DataFrame({'src_node': edges.index.get_level_values(1),
                           'dest_node': edges.values})
-    edges['x0'] = [pts.loc[n].x for n in edges.src_node]
-    edges['x1'] = [pts.loc[n].x for n in edges.dest_node]
-    edges['y0'] = [pts.loc[n].y for n in edges.src_node]
-    edges['y1'] = [pts.loc[n].y for n in edges.dest_node]
+
+    edge_list = []
+    for i in edges.index:
```
qiyunzhu: I wonder if this is the most efficient way to loop thru a data frame. Maybe extract the columns into Python lists then loop will be faster?

mortonjt: This should be ok - looping through a data frame this way is just as fast.  I can send over benchmarks if necessary.

PR #164, opened 2017-04-21
MAINT: update travis for new QIIME 2 automatic builds

ebolyen:
This should let travis pass now. It's using the new environment file generated by QIIME 2's automatic CI system, which creates new conda packages for every merged commit.

I'm not completely sure how you want to organize things here, but this can be used as a template if nothing else.


mortonjt:
Thanks @ebolyen ! You are the man!

PR #166, opened 2017-04-23
STY: Refactoring OLS summary

mortonjt:
1. The plots have been condensed into a single multipanel figure
2. Radial plots are mutable
3. Heatmap summary of the regression plot is added

TODO: Need more tests


mortonjt:
Thanks!

PR #172, opened 2017-05-01
Balance summary

mortonjt:
Addresses https://github.com/biocore/gneiss/issues/181

Still a work in progress - need to figure out how to fix the scaling in the rendering.

This basically provides some better summaries for the individual balances. This ultimately allows for users to visualize how a single balance relates to the metadata and what microbes the balance is composed of.

Below is an example how to run it and the results of the visualization.

```
qiime gneiss balance-taxonomy \
    --i-model 88soils_regression_model.qza \
    --o-visualization 88soils_summary
```

screen shot 2017-05-27 at 7 46 35 pm

Help menu
```
tests-MacBook-Pro-4:tests mortonjt$ qiime gneiss balance-taxonomy
Usage: qiime gneiss balance-taxonomy [OPTIONS]

  Visualize the distribution of a single balance and summarize its numerator
  and denominator components.

Options:
  --i-balances PATH               Artifact: FeatureTable[Balance]  [required]
                                  The table of balances resulting from the ilr
                                  transform.
  --i-tree PATH                   Artifact: Phylogeny[Rooted]  [required]
                                  The
                                  tree used to calculate the balances.
  --i-taxonomy PATH               Artifact: FeatureData[Taxonomy]  [required]
                                  Taxonomy information for the OTUs.
  --p-balance-name TEXT           [required]
                                  Name of the balance to summarize.
  --p-taxa-level [family|genus|kingdom|class|order|species|phyla]
                                  [default: phyla]
                                  Level of taxonomy to
                                  summarize.
  --m-metadata-file PATH          Metadata mapping file  [optional]
  --m-metadata-category TEXT      Category from metadata mapping file
                                  [optional]
  --o-visualization PATH          Artifact: Visualization  [required if not
                                  passing --output-dir]
  --output-dir DIRECTORY          Output unspecified results to a directory
  --cmd-config PATH               Use config file for command options
  --verbose                       Display verbose output to stdout and/or
                                  stderr during execution of this action.
                                  [default: False]
  --help                          Show this message and exit.
```
@nbokulich do you have any thoughts on the user interface here?


nbokulich:
I'd recommend making `taxa_level` an `int` as levels may change depending on the taxonomy.

Also, don't place an upper limit, as not all taxonomies will necessarily have 6 or 7 tidy levels. (if `taxa_level` > len(taxon): `taxa_level` = taxon[-1])

Looks great!

mortonjt:
Finally this passed!  @qiyunzhu do you mind if you could take a look at this?

mortonjt:
@nbokulich @ebolyen do you know how to force the ordering of options on the command line?
For instance here, it would be great if we could force the ordering `['kingdom', 'phylum', 'class', 'order', 'family', 'genus', species']`.

ebolyen:
@mortonjt the underlying choices are stored as a set, so there's not a way to do that at this time.

gneiss/plot/tests/test_plot.py
```diff
@@ -87,5 +88,131 @@ def test_visualization_small(self):
                           html)
 
 
+class TestBalanceTaxonomy(unittest.TestCase):
+
+    def setUp(self):
+        self.results = "results"
+        os.mkdir(self.results)
```
qiyunzhu: Will it cause problem without making a temporary directory?

gneiss/plot/_plot.py
```diff
@@ -8,17 +8,130 @@
...
+        index_f.write('<html><body>\n')
+        if metadata is not None:
+            index_f.write('<h1>Balance vs %s </h1>\n' % c.name)
+            index_f.write(('<img src="balance_metadata.svg" '
+                           'alt="barplots">\n\n'))
+
+        index_f.write('<h1>Balance Taxonomy</h1>\n')
+        index_f.write('<img src="barplots.svg" alt="barplots">\n\n')
+        index_f.write(('<h3>Numerator taxa</h3>\n'
+                       '<a href="numerator.csv">\n'
```
qiyunzhu: Trivial. There two HTML lines can be merged in one.

gneiss/plot/_decompose.py
```diff
@@ -111,22 +113,35 @@ def balance_barplots(tree, balance_name, header, feature_metadata,
...
         f, (ax_num, ax_denom) = plt.subplots(2)
     else:
         ax_num, ax_denom = axes[0], axes[1]
     st = tree.find(balance_name)
-    num = feature_metadata.loc[list(st.children[0].subset())]
-    denom = feature_metadata.loc[list(st.children[1].subset())]
+    num_clade = st.children[NUMERATOR]
+    denom_clade = st.children[DENOMINATOR]
+    if num_clade.is_tip():
+        num_ = pd.DataFrame(
```
qiyunzhu: This `num_` variable is not declared before the `if` block?

mortonjt: That doesn't matter - since it is declared in both the if and the else.

gneiss/plot/_plot.py
mortonjt: bam done!

gneiss/plot/tests/test_plot.py
mortonjt: Nope - because I tear it down right after creation.  The main reason why I do this is because it makes debugging a heck of a lot easier.

It is making the assumption that the current directory is clean and doesn't have the folder `results`.  This is a safe assumption, particularly given that these file directories are created on the fly.

mortonjt:
1) The NUMERATOR and DENOMINATOR can't be parameters -- those quantities are fixed constants.  Basically, this is performing a mapping such that the right child maps to the numerator of the balance, and the left child maps to the denominator of the balance.  These constants are crucial for making sure that this isn't mixed up (it is very surprisingly easy to mix numerator with denominator).
2) Most definitely - we would like to have this generalized to handle feature-metadata in general.  However, there are multiple structural obstacles for making this happen.  Namely a fixed file format for accepting this sort of object and q2 compatibility.  Right now, there doesn't exist a feature-metadata object in qiime2, but there is a `FeatureData[Taxonomy]` object.  Right now, we are only concerning the `FeatureData[Taxonomy]` object, which only has taxonomy.  I am not familiar with taxa strings with more than 7 levels, or what the semantics of that would look like.  @nbokulich @antgonza do you have an example readily available?

nbokulich:
An example of a taxonomy with > 7 levels is the raw SILVA taxonomy. It has the same format as greengenes but contains more levels, e.g.:
```
AB001038.1.1721	Eukaryota;Archaeplastida;Chloroplastida;Chlorophyta;Chlorophyceae;Chlamydomonadales;Polytoma;Chlamydomonas pulsatilla
```
But there are other taxonomies that contain fewer than 7 levels, e.g., RDP taxonomy I believe is 6 levels.

Another example (to consider non-taxonomic feature data) would be gene pathway/ontology data. E.g., picrust data reports KEGG pathways that are 3 levels deep.

mortonjt:
Ok!  I have made the taxonomies integer valued, so that multiple levels can be accessed.

PR #173, opened 2017-05-02
ENH: Assign ids

mortonjt:
This creates UUIDs for internal nodes as @wasade suggested.

This will allow for trees to be readily read into gneiss.


gneiss/cluster/_cluster.py
```diff
@@ -121,3 +123,31 @@ def gradient_clustering(table: pd.DataFrame,
...
                  'the underlying block-like structures.')
 )
+
+
+def assign_ids(tree: skbio.TreeNode) -> skbio.TreeNode:
+    def replace_dashes(x):
+        return str(x).replace('-', '')
+    t = tree.copy()
+    t.bifurcate()
+    ids = ['%sL_%s' % (i, replace_dashes(uuid.uuid4()))
```
wasade: if you do this, you can't test to see if the id is a uuid. are you sure you want to replace the dashes?

mortonjt:
The main reason I'm thinking about doing this is so that we don't restrict
operations when these are loaded into pandas. I don't have strong feelings
here - we can switch back if you believe that there is loss of major
functionality

On May 1, 2017 6:58 PM, "Daniel McDonald" <notifications@github.com> wrote:

*@wasade* commented on this pull request.
------------------------------

In gneiss/cluster/_cluster.py
<https://github.com/biocore/gneiss/pull/173#discussion_r114236841>:

> @@ -121,3 +123,31 @@ def gradient_clustering(table: pd.DataFrame,
                  'This method is primarily used to sort the table to reveal '
                  'the underlying block-like structures.')
 )
+
+
+def assign_ids(tree: skbio.TreeNode) -> skbio.TreeNode:
+    def replace_dashes(x):
+        return str(x).replace('-', '')
+    t = tree.copy()
+    t.bifurcate()
+    ids = ['%sL_%s' % (i, replace_dashes(uuid.uuid4()))

if you do this, you can't test to see if the id is a uuid. are you sure you
want to replace the dashes?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/biocore/gneiss/pull/173#pullrequestreview-35686100>,
or mute
the thread

mortonjt:
It can,  but some operations are limited (i.e. you can do. df
['your_uuid'], but not df.your_uuid)

Some of the indexing options won't be available,  but I think that is ok.
The dashes are back

tkosciol:
Looks good. I don't have write permissions to merge, though.

mortonjt:
Thanks!

PR #175, opened 2017-05-04
WIP: Labeled dendrogram

mortonjt:
Work in progress.  This creates dendrograms with labels.  Mainly useful for debugging small trees.


mortonjt:
All visualizations are going to be pushed into empress

PR #179, opened 2017-05-08
URGENT: Updating types in light of q2-composition

mortonjt:
In light of the PR merged here: https://github.com/qiime2/q2-composition/pull/34

The Balance type that was merged [earlier](https://github.com/biocore/gneiss/pull/150) will need to be removed to remove the type clashing.


antgonza:
👍 , when tests pass but it looks like a lot of commits for the amount of changes, is that OK?

mortonjt:
Yes - but it would be great it this could be squashed.  There was quite a few commits to revert changes.

mortonjt:
Thanks @antgonza !

PR #174, opened 2017-05-03
WIP: Major refactor of regression object

mortonjt:
This refactor is unrolling the preprocessing.  So now, a table of balances can be pass directly into the regression models.

Here's the list of things that needs to get review/merged before this can get merged.

- [x] (1) q2-cli regression refactor - https://github.com/mortonjt/gneiss/pull/10
- [x] (2) Regression summaries - https://github.com/mortonjt/gneiss/pull/9
- [x] (3) Linear mixed effects refactor - https://github.com/mortonjt/gneiss/pull/7

PR #182, opened 2017-05-12
BUG: Fixing Ilr fix

mortonjt:
There was an issue, where the tree wasn't properly aligned to the table when the ilr transform was applied (i.e. the tip names in the tree wasn't appropriately matched to the column names in the table).  So the ilr transform would be botched.

Note that this was only an issue for the new `ilr-transform` command in the CLI.

@antgonza do you mind if you could take a look at this PR?  The changes are very minimal


antgonza:
looks good to me, 👍  when tests pass.

mortonjt:
Looks like the errors are completely unrelated to this PR.

I'm currently cleaning up another PR in #174 to get rid of the q2 types that are throwing these errors.
So we'll probably want to merge that first.  Note that it has already been reviewed - so its just a matter of getting the tests to pass.

antgonza:
note that this will break master, what about creating a new branch, where both prs are deployed?

mortonjt:
I'm going to close this, since the fixes have been translated to #174

PR #186, opened 2017-05-27
WIP: OLS CV speedup

mortonjt:
This is a working progress, so there are a few kinks that need to be worked out. But the bulk of the work is in place.

@wasade, this is aiming to address some of the scalability issues that we are seeing.  This PR is aiming to get around those issues by removing the intermediate statsmodels objects and doing the raw matrix multiplications.  Nice this about this, we can distribute matrix multiplication across nodes via dask.  However, note that this PR by itself was able to crank through the AG results within 30min, pulling off a 100x speedup.

I have been able to test this against statsmodels and am able to get identical results.  In addition, this major refactor was able to undercover some sneaky bugs in the cross validation.

There are some specific concerns that I would like to have some feedback on - I'll be listing them within the comments.


gneiss/regression/_ols.py
```diff
@@ -185,17 +177,74 @@ class OLSModel(RegressionModel):
...
+        Y = self.response_matrix
+        X = self.design_matrices
+        n, p = X.shape
+        inv = np.linalg.pinv(np.dot(X.T, X))
+        cross = np.dot(inv, X.T)
+        beta = np.dot(cross, Y)
+        pX = np.dot(X, beta)
+        resid = (Y - pX)
+        sst = (Y - Y.mean(axis=0))
+        sse = (resid**2).sum(axis=0)
```
mortonjt: @wasade just as a heads up, all of the bulk computations are being done in this block.  Note that I have not done scalability tests on this yet.  For now, parallelization can be utilized with numpy + openmp.

mortonjt: @serenejiang I would like your feedback here - do these matrix equations make sense?  Here we are trying to solve the least squares minimization problem.

gneiss/regression/_ols.py
```diff
@@ -288,36 +314,34 @@ def kfold(self, num_folds=10, **kwargs):
...
             model_mse = np.mean(model_resid.sum(axis=0))
             results.loc['fold_%d' % k, 'model_mse'] = model_mse
             results.loc['fold_%d' % k, 'Rsquared'] = res_i.r2
 
             # prediction error
-            predicted = res_i.predict(exog.iloc[test])
-            pred_resid = ((predicted - endog.iloc[test])**2)
+            p = res_i.predict(X=self.design_matrix.iloc[test]).values
+            r = self.response_matrix.iloc[test].values
+            pred_resid = ((p - r)**2)
```
mortonjt: These lines I'm not completely sure about.  I'm trying to calculate mse for both the model error and the prediction error.  The degrees of freedom may be a little off - @serenejiang do you have thoughts here wrt `pred_mse` and `model_mse`?

gneiss/regression/_ols.py
```diff
@@ -407,37 +431,27 @@ def lovo(self, **kwargs):
...
+            res_i.fit(**kwargs)
+            predicted = res_i.predict()
+            r = self.response_matrix.values
+            p = predicted.values
+
+            model_resid = ((r - p)**2)
+            model_mse = np.mean(model_resid.sum(axis=0))
+            results.loc[feature_id, 'mse'] = model_mse
+            results.loc[feature_id, 'Rsquared'] = res_i.r2
+            results.loc[feature_id, 'R2diff'] = self.r2 - res_i.r2
```
mortonjt: This is trying to do some relative importance analysis - basically trying to see how much the R^2 decreases when a specific variable is left out.  @serenejiang do you have any thoughts here?

gneiss/regression/_mixedlm.py
```diff
@@ -145,7 +147,11 @@ def mixedlm(formula, table, metadata, groups, **kwargs):
                           **kwargs)
         submodels.append(mdf)
 
-    return LMEModel(submodels, balances=table)
+    # ugly hack to get around the statsmodels object
```
mortonjt: In case you are wondering why we are doing this I raised an issue here #187.
Basically, the multidimensional response statistics in statsmodels are lacking at the moment, so we are doing some hacks here to overcome that.

gneiss/_model.py
```diff
@@ -7,29 +7,28 @@
...
-            balances are columns. These balances were calculated
-            using `tree`.
+        Y : pd.DataFrame
+            Response matrix.  This is the matrix being predicted.
+            Also known as the dependent variable in univariate analysis.
+        Xs : iterable of pd.DataFrame
+            Design matrices.  Also known as the independent variable
+            in univariate analysis. Note that this allows for multiple
+            design matrices to be inputted to enable multiple data block
+            analysis.
```
mortonjt: In case you are wondering why refactor this section.
We want to ultimately be able to enable multiblock regression / classification #188
So it would be great if we could pass in a single response matrix, and multiple design matrices to try to predict the regression matrix.

gneiss/_model.py
```diff
@@ -80,3 +79,10 @@ def write_pickle(self, filename):
                 pickle.dump(self, fh)
         else:
             pickle.dump(self, filename)
+
+    def percent_explained(self):
+        """ Proportion explained by each principal balance."""
+        # Using sum of squares error calculation (df=1)
+        # instead of population variance (df=0).
+        axis_vars = np.var(self.response_matrix, ddof=1, axis=0)
+        return axis_vars / axis_vars.sum()
```
mortonjt: Bumped this up here, because this is relatively agnostic of the underlying model.

gneiss/regression/_mixedlm.py
```diff
@@ -221,3 +220,132 @@ def percent_explained(self):
...
+            coef = coef.append(c)
+
+        if tree is not None:
+            basis, _ = balance_basis(tree)
+            c = ilr_inv(coef.values.T, basis=basis).T
+
+            return pd.DataFrame(c, index=[n.name for n in tree.tips()],
+                                columns=coef.columns)
+        else:
+            return coef.T
```
mortonjt: This nasty code is a result of this #187
If we can code up the multivariate linear mixed effects, we don't need statsmodels and can seriously clean up / speed up this code.

gneiss/plot/tests/test_regression_plot.py
```diff
@@ -63,111 +63,15 @@ def test_visualization(self):
...
-        npt.assert_allclose(exp_pred.values, pred.values,
-                            rtol=1e-2, atol=1e-2)
+        fp = os.path.join(self.results, 'pvalues.csv')
+        self.assertTrue(os.path.exists(fp))
+        fp = os.path.join(self.results, 'coefficients.csv')
+        self.assertTrue(os.path.exists(fp))
+        fp = os.path.join(self.results, 'predicted.csv')
+        self.assertTrue(os.path.exists(fp))
+        fp = os.path.join(self.results, 'residuals.csv')
+        self.assertTrue(os.path.exists(fp))
```
mortonjt: Note, these values were changed for some reason - will require some investigation

gneiss/regression/_ols.py
serenejiang: Are you treating each of the response variables separately in the least squares estimation? If yes, then your matrix equation is good. Or are you aiming at the multivariate response least squares estimation?

mortonjt:
This is actually a multivariate response regression. So the Y here is high
dimensional.  Note that we are seeing sane fits with promising cross
validation results when running this.

gneiss/regression/_ols.py
mortonjt: This is a multivariate response least squares estimation.  But I have worked out the examples - this is identical to running OLS on the response variables separately - I have tests under test_ols.py to prove this.

gneiss/plot/tests/test_regression_plot.py
mortonjt: I think I found the culprit. 

The previous test cases were designed to be integration tests -- so we are likely to be seeing some rounding errors here.

I'm going to actually keep these test cases removed, since we are already rigorously testing this in test_ols.py

gneiss/_model.py
```diff
@@ -40,43 +38,9 @@ def summary(self):
...
-            Output file to store pickled object.
-        """
-        if isinstance(filename, str):
-            with open(filename, 'wb') as fh:
-                pickle.dump(self, fh)
-        else:
-            pickle.dump(self, filename)
+    def percent_explained(self):
+        """ Proportion explained by each principal balance."""
+        # Using sum of squares error calculation (df=1)
```
antgonza: Not sure this is doing what you expect, would you mind explaining a bit more. My concern is about the squares error calculation vs. what the ddof does, from the manual: “Delta Degrees of Freedom”: the divisor used in the calculation is N - ddof, where N represents the number of elements. By default ddof is zero.

gneiss/plot/tests/test_regression_plot.py
antgonza: Is it concerning that even with assert_allclose the numbers don't match? I'm guessing that it's the same input, right?

gneiss/plot/tests/test_regression_plot.py
```diff
@@ -63,111 +63,15 @@ def test_visualization(self):
...
-                   'y1': 0.31931185594151867},
-            'x2': {'y0': 0.31929793802591028,
-                   'y1': 0.3192984932257481},
-            'x3': {'y0': 0.31930876472902192,
-                   'y1': 0.31930931993001832},
-            'x4': {'y0': 0.31931786743864193,
-                   'y1': 0.31931842264061172}})
-
-        npt.assert_allclose(exp_pvals.values, pvals.values,
-                            rtol=1e-2, atol=1e-2)
```
antgonza: Same here ... about results not being close

gneiss/regression/_mixedlm.py
```diff
@@ -221,3 +220,132 @@ def percent_explained(self):
         # instead of population variance (df=0).
         axis_vars = np.var(self.balances, ddof=1, axis=0)
         return axis_vars / axis_vars.sum()
+
+    def coefficients(self, tree=None):
+        """ Returns coefficients from fit.
+        Parameters
```
antgonza: new line?

gneiss/regression/_model.py
```diff
@@ -128,36 +131,25 @@ def predict(self, X=None, tree=None, **kwargs):
...
-            # check if X is none.
-            p = pd.Series(m.predict(X, **kwargs))
-            p.name = m.model.endog_names
-            if X is not None:
-                p.index = X.index
-            else:
-                p.index = m.fittedvalues.index
-            prediction = prediction.append(p)
-
+        pass
```
antgonza: Do we need this pass? It will avoid execution of the rest of the code, right?

gneiss/regression/_ols.py
```diff
@@ -192,10 +185,110 @@ def fit(self, **kwargs):
         ----------
         **kwargs : dict
            Keyword arguments used to tune the parameter estimation.
+        """
```
antgonza: I think it will be worth having some documentation of what we are doing on each step so we can follow the description and code.

gneiss/regression/tests/test_ols.py
```diff
@@ -15,334 +15,234 @@
...
-        }, index=['s1', 's2', 's3', 's4', 's5',
-                  's6', 's7', 's8', 's9', 's10',
-                  's11', 's12', 's13', 's14', 's15'])
-
     def tearDown(self):
         shutil.rmtree(self.results)
 
+    def test_ols_test(self):
 
-class TestOLSFunctions(TestOLS):
```
antgonza: Are all these test really not needed? Where did these test move? For example, we don't have test_ols or test_ols_table_immutable in the new code ...

gneiss/_model.py
mortonjt: Sure thing!

The degrees of freedom boils down to the nature of [variance](https://en.wikipedia.org/wiki/Variance#Population_variance_and_sample_variance).
When calculating population variance, you are just doing

`sigma = (1/n)* sum([(y[i] - np.mean(y))**2 for i in range(0, n)])`

The problem is, this is actually a biased estimator for calculating variance - it is off by 1 degree of freedom.  The way to obtain an unbiased estimator of variance is

`sigma = (1 / (n+1))* sum([(y[i] - np.mean(y))**2 for i in range(0, n)])`

so by specifying `ddof=1`, I am forcing numpy to calculate the sample variance - because by default, numpy is calculating the biased estimate of variance.  Super subtle, but super important.

gneiss/regression/_model.py
mortonjt: whoa ... that needs to be removed ...

gneiss/plot/tests/test_regression_plot.py
mortonjt: I think we are dealing with an inherently bad unittest.  And we have been having problems with these unittests for a long time -- the precision of the unittests have been altered in previous iterations.

For example, here is the following error when running this test.
```
(mismatch 86.66666666666667%)
 x: array([[-0.533751, -0.564799, -0.562873, -0.541892, -0.513249, -0.485805,
        -0.465883, -0.457265, -0.461186, -0.476321, -0.498785, -0.522126,
        -0.537322, -0.532768, -0.494272],...
 y: array([[-0.569111, -0.549032, -0.531633, -0.516916, -0.50488 , -0.495526,
        -0.488852, -0.48486 , -0.483549, -0.484919, -0.488971, -0.495703,
        -0.505117, -0.517212, -0.531989],...
```
So we are most likely observing rounding errors here, because the majority of the values in the matrix are correct.

The reason why I am not worried about removing these particular checks is because we have much more robust / easily verifiable tests in the `test_ols.py` -- I am very confident based on these tests that these modules are working.

gneiss/_model.py
antgonza: Thanks, my source of confusion is that this line says calculate squares error calculation vs. sample variance ... not sure if that's the same or not

gneiss/plot/tests/test_regression_plot.py
antgonza: A mismatch of ~87% is not close, right? Anyway, if you are confident and we also have this kind of tests (making sure that the same input gives the same output) in test_ols.py, I'm OK with this.

gneiss/regression/tests/test_ols.py
mortonjt: Yeah ... I really need to include those back in.  Thanks for catching that!

mortonjt:
Thanks for the feedback @antgonza @serenejiang !  Now we do have a few smaller maintenance PRs coming in tomorrow - so let me know by then if you guys spot any major additional problems in this PR by then.  Thanks!

PR #194, opened 2017-06-13
BLD: Fix setup.py extra_require -> extras_require

ElDeveloper:
There was a typo in setup.py that prevented users from installing
gneiss' QIIME2 dependencies.  This also adds a line in needed
.travis.yml to install gneiss using the [q2] extras.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
   - conda install --yes -n test_env cython
   - source activate test_env
   - pip install -r ci/pip_requirements.txt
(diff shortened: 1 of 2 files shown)


mortonjt:
Thanks for submitting this @ElDeveloper !  The error that you are seeing is coming from the newest metadata object in qiime2.  That is actually being fixed in #186 - so we can merge this immediately after merging that PR.

mortonjt:
Thanks!

PR #197, opened 2017-06-16
Documentation cleanup

mortonjt:


mortonjt:
We'll go ahead and merge this in so that we can cut out a conda recipe today.  We'll be pushing out another PR to fix the heatmaps that are addressed by #200 in a bit.

PR #201, opened 2017-06-17
Updating heatmaps in notebook

mortonjt:


mortonjt:
The online docs look good, so we're merging in and cutting the beta release.

PR #203, opened 2017-06-17
ENH: Changing types over to Hierarchy instead of Phylogeny[Rooted]

mortonjt:
This addresses #202


mortonjt:
This is a small fix - so we're going to merge this to finalize the notebooks.

Issue #193, opened 2017-06-13
Decoupled tutorial or QIIME1-based tutorial?


adityabandla:
Hi Jamie

I have had a frustrating time since the weekend getting GNEISS to work with QIIME2. Can GNEISS be delinked from QIIME2?

Best,
Aditya

mortonjt:
I'm putting this on the milestone for version 1.  Once we have the qiime2 / gneiss APIs in a more stable condition, it'll be more straightforward on how to do this.

mortonjt:
This has been fixed in #214 .  Thanks for raising this @adityabandla !  You should see these changes in the next release once the conda release is out.

PR #214, opened 2017-07-09
Q2 plugin refactor

mortonjt:
This addresses #193 and decouples the qiime2 plugin from the Python API

The q2 plugin can be found [here](https://github.com/mortonjt/q2-gneiss)


antgonza:
Code looks fine but there are errors ...

mortonjt:
@antgonza given that this PR is passing, would you mind that you could confirm that this is good to go?

mortonjt:
Thanks!

PR #234, opened 2017-08-17
Proportion plots

mortonjt:
This will summarize the mean proportions for each of the balances

You can use this module as follows

```python
import pandas as pd
from gneiss.plot._decompose import proportion_plot

table = pd.DataFrame({
    'A': [1, 1.2, 1.1, 2.1, 2.2, 2],
    'B': [9.9, 10, 10.1, 2, 2.4, 2.1],
    'C': [5, 3, 1, 2, 2, 3],
    'D': [5, 5, 5, 5, 5, 5],
}, index=['S1', 'S2', 'S3', 'S4', 'S5', 'S6'])

feature_metadata = pd.DataFrame({
    'A': ['k__foo', 'p__bar', 'c__', 'o__', 'f__', 'g__', 's__'],
    'B': ['k__foo', 'p__bar', 'c__', 'o__', 'f__', 'g__', 's__'],
    'C': ['k__poo', 'p__tar', 'c__', 'o__', 'f__', 'g__', 's__'],
    'D': ['k__poo', 'p__far', 'c__', 'o__', 'f__', 'g__', 's__']
}, index=['kingdom', 'phylum', 'class', 'order', 'family', 'genus', 'species']).T

metadata = pd.DataFrame({'groups': ['X', 'X', 'X', 'Y', 'Y', 'Y'],
                         'dry': [1, 2, 3, 4, 5, 6]}, 
                        index=['S1', 'S2', 'S3', 'S4', 'S5', 'S6'])

num_features = ['A', 'B']
denom_features = ['C', 'D']
ax1, ax2 = proportion_plot(table, metadata, num_features, denom_features, 
                           feature_metadata, 'groups', 'X', 'Y', taxa_level='phylum')
```

image

@wasade @cuttlefishh do you have any thoughts on the current API?  Do you think it is intuitive, or should this be simplified?

### DIFF
--- a/gneiss/plot/_decompose.py
+++ b/gneiss/plot/_decompose.py
@@ -160,7 +160,7 @@ def balance_barplots(tree, balance_name, header, feature_metadata,
 def proportion_plot(table, metadata, category, left_group, right_group,
                     num_features, denom_features,
                     feature_metadata=None,
-                    aggregate_level='species',
+                    label_col='species',
                     num_color='#105d33', denom_color='#b0d78e',
                     axes=(None, None)):
     """ Plot the mean proportions of features within a balance.
@@ -185,12 +185,19 @@ def proportion_plot(table, metadata, category, left_group, right_group,
     right_group : str
        Name of group within sample metadata category to plot
        on the right of the plot.
-    aggregrate_col : str
+    label_col : str
        Column in the feature metadata table to summarize by.
     num_color : str
        Color to plot the numerator.
     denom_color : str
        Color to plot the denominator.
+
+    Returns
+    -------
+    ax_num : matplotlib.pyplot.Axes
+       Matplotlib axes for the numerator bars
+    ax_denom : matplotlib.pyplot.Axes
+       Matplotlib axes for the denominator bars
     """
     import seaborn as sns
     if axes[0] is None or axes[1] is None:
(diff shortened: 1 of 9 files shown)


gneiss/plot/_decompose.py
```diff
@@ -155,3 +155,137 @@ def balance_barplots(tree, balance_name, header, feature_metadata,
...
+                    label_col='species',
+                    num_color='#105d33', denom_color='#b0d78e',
+                    axes=(None, None)):
+    """ Plot the mean proportions of features within a balance.
+
+    This plots the numerator and denominator components within a balance.
+
+    Parameters
+    ----------
+    table : pd.DataFrame
```
wasade: are you concerned about requiring dense matrices for the table data?

mortonjt: Nope

cuttlefishh:
The API looks pretty good to me. For the plots, I have some suggestions to make them clearer:

* Change x-axis from "proportion -- proportion" to "Denominator proportion -- Numerator proportion".  Alternatively, give the user the option to edit these axis labels manually. Title labels should be editable too.
* Would it be useful for the y-axis labels (currently taxa) to be more flexible? I'm thinking these could be OTU identifiers instead of (or in addition to) taxon names. OTU identifiers make more sense for reference-based methods, but one could imagine assigning identifiers to Deblur sequences and wanting to use those.

The more customizable the plots are, the better people can get what they want, and the more likely they are to use the plots.

mortonjt:
Thanks for taking a look over it Luke!

The user actually already has full control to modify the underlying plots -- this function is returning a full matplotlib figure.  But maybe it would be useful to have some clear cut examples in the documentation about how to modify these labels.  What do you think?

mortonjt:
Since the comments have been addressed, and this PR has been outstanding for a while, I'm going to go ahead and merge this in so that we can start integrating this functionality into the q2-gneiss plugin.

PR #250, opened 2018-01-28
Sparse ilr

mortonjt:
This creates a sparse representation of an ilr basis.  This will ultimately enable scalability of very high dimensional datasets.

This will replace the efforts in #8


gneiss/balances.py
```diff
@@ -52,7 +54,7 @@ def _balance_basis(tree_node):
         basis[i, :] = np.array([0]*k[i] + [a[i]]*r[i] + [b[i]]*s[i] + [0]*t[i])
     # Make sure that the basis is in level order
     basis = basis[:, ::-1]
-    nds = list(nds)
+    nds = [n.name for n in nds]
```
mortonjt: Note that we are only keep track of names instead of entire TreeNode objects.  This is to ensure consistency between the methods and immutability.

gneiss/balances.py
```diff
@@ -162,3 +164,99 @@ def _count_matrix(treenode):
...
+        Input bifurcating tree.  Must be strictly bifurcating
+        (i.e. every internal node needs to have exactly 2 children).
+        This is used to specify the ilr basis.
+
+    Returns
+    -------
+    scipy.sparse.coo_matrix
+       The ilr basis required to perform the ilr_inv transform.
+       This is also known as the sequential binary partition.
+       Note that this matrix is represented in clr coordinates.
```
mortonjt: Note that this is in clr coordinates, which is the same as the `_balance_basis` function.

mortonjt:
We have quite a few outstanding PRs, so let's merge this in once travis completes.

PR #253, opened 2018-02-12
Sparse utils

mortonjt:
This adds support for sparse data manipulation, namely

- Aligning trees against biom tables
- Sparse indexing of dense matrix multiplication
- Minibatch selection

Depends on #250


mortonjt:
There are some really weird errors going on with nose, making it difficult to get this functionality in.  
@antgonza , @ElDeveloper any idea what is going on?  I'm able to produce this on my laptop, but not sure why nose is behaving this way.

ElDeveloper:
From the looks of the travis output, the problem does not seem to be coming from nose, it seems to becoming from the coverage tool. It appears to be interpreting that `--with-doctest` is an argument for setup.py. Later on setup.py errors as that's not an available option. I would try changing the makefile to enclose `nose --with-doctest` in single-quotes or use a double-hyphen to separate the  positional arguments i.e `..... setup.py -- nose --with-doctest`. If that doesn't work then I don't have any other ideas. ` ¯\_(ツ)_/¯`

mortonjt:
Alright - it looks like the tests are passing.  If there aren't any issues with this PR, I'll be merging this in over the next few days.

What do you think @antgonza ?

antgonza:
Wanna add test so the coverage doesn't decrease?

mortonjt:
@antgonza thanks for catching that -- there were actually a couple of nasty bugs that would have slipped through if it weren't for that test.  It should be ok now.

antgonza:
Kewl, any idea why coveralls didn't run this time?

PR #255, opened 2018-02-13
WIP: Poisson regression

mortonjt:
Adding in new poisson regression model.

Depends on #250 #253 and #254 

Note that this isn't finalized, since we may be swapping out the MAP estimate with Variational Bayes


mortonjt:
Closing since we will not be using Edward.

PR #257, opened 2018-02-25
FIX: Adding in option to download FDR pvalues

mortonjt:
Adding in option to download FDR pvalues as requested [here](https://forum.qiime2.org/t/gneiss-questions/1284/9) and [here](https://forum.qiime2.org/t/adding-download-option-for-fdr-adjusted-p-value-table-in-gneiss/2995)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # gneiss changelog
 
 ## Version 0.4.2-dev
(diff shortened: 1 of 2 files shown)


mortonjt:
fyi @thermokarst and @ebolyen this now enables the download of FDR corrected pvalues and is good to go.

PR #265, opened 2018-08-21
Update _composition.py

lisa55asil:


mortonjt:
Looks good! Could you move it inside of the function past the def? That'll fix the Travis errors

mortonjt:
This is now fixed @lisa55asil  thanks!

Issue #270, opened 2019-01-09
Sign change in ilr_transform results using new dependencies


thermokarst:
The test suite currently fails:

```bash
____________________________________________ TestILRTransform.test_ilr _____________________________________________
E       AssertionError: DataFrame.iloc[:, 1] are different
E
E       DataFrame.iloc[:, 1] values are different (33.33333 %)
E       [left]:  [-5.551115123125783e-17, 0.49012907173427367, 5.551115123125783e-17]
E       [right]: [-5.551115e-17, -0.4901291, 5.551115e-17]
```

with the following deps:
- `numpy 1.15.2`
- `scipy 1.1.0`
- `pandas 0.23.4`

We are seeing a [similar failure](https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/unit-test-q2-gneiss/builds/10) in the test suite for q2-gneiss.

<details>
  <summary>Full test log</summary>

```bash
pytest gneiss                                                                                     master 2bed3e3
=============================================== test session starts ================================================
platform darwin -- Python 3.5.5, pytest-3.8.1, py-1.7.0, pluggy-0.8.0
rootdir: /Users/matthew/Desktop/gneiss, inifile:
collected 113 items

gneiss/cluster/tests/test_pba.py .....                                                                       [  4%]
gneiss/composition/tests/test_composition.py F                                                               [  5%]
gneiss/composition/tests/test_variance.py ..                                                                 [  7%]
gneiss/plot/tests/test_decompose.py .......                                                                  [ 13%]
gneiss/plot/tests/test_dendrogram.py ........                                                                [ 20%]
gneiss/plot/tests/test_heatmap.py ....                                                                       [ 23%]
gneiss/plot/tests/test_radial.py .                                                                           [ 24%]
gneiss/plot/tests/test_regression_plot.py ..                                                                 [ 26%]
gneiss/regression/tests/test_mixedlm.py ...                                                                  [ 29%]
gneiss/regression/tests/test_model.py .....                                                                  [ 33%]
gneiss/regression/tests/test_ols.py .............                                                            [ 45%]
gneiss/tests/test_balances.py .............                                                                  [ 56%]
gneiss/tests/test_model.py .                                                                                 [ 57%]
gneiss/tests/test_sort.py ................                                                                   [ 71%]
gneiss/tests/test_util.py ................................                                                   [100%]

===================================================== FAILURES =====================================================
____________________________________________ TestILRTransform.test_ilr _____________________________________________

[...]

PR #272, opened 2019-01-09
Addressing #270

mortonjt:
Bumping up skbio version
Simplifying ilr test by hard coding the tree

### DIFF
--- a/gneiss/composition/tests/test_composition.py
+++ b/gneiss/composition/tests/test_composition.py
@@ -9,8 +9,8 @@ import unittest
 import numpy as np
 import pandas as pd
 from gneiss.composition._composition import ilr_transform
-from gneiss.cluster import gradient_linkage
 import pandas.util.testing as pdt
+from skbio import TreeNode
 
 
 class TestILRTransform(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


ci/conda_requirements.txt
```diff
@@ -4,7 +4,7 @@ pycodestyle
 flake8
 IPython>4.0.0
 notebook
-scikit-bio=0.5.1
+scikit-bio=0.5.5
```
thermokarst: Maybe this should be a minimum pin, instead of an absolute pin, like what you did below in `setup.py`?

gneiss/composition/tests/test_composition.py
```diff
@@ -23,15 +23,19 @@ def test_ilr(self):
                              index=[1, 2, 3],
                              columns=['a', 'b', 'c', 'd'])
         table = table.reindex(columns=np.random.permutation(table.columns))
-        ph = pd.Series([1, 2, 3], index=table.index)
-        tree = gradient_linkage(table, ph)
+        # fix tree
+        tree = TreeNode.read([r"((d,c)y2, (b,a)y1)y0;"])
```
thermokarst: Is this problem really just in the test suite? Seems like the underlying commands are producing different results, from what I can tell...

mortonjt:
The tests are passing once the tree is fixed, but there is something a bit funky going on with python3.6. @thermokarst is qiime2 still only running python3.5?

thermokarst:
> @thermokarst is qiime2 still only running python3.5?

We are currently exploring upgrading the minimum runtime to python 3.7 right now - we are planning on writing up something about this on the QIIME 2 Forum soon (maybe beginning of next week?)

ci/conda_requirements.txt
mortonjt: nice catch! Yes, that should have been a min pin

gneiss/composition/tests/test_composition.py
mortonjt: found the problem - turns out that any operations you perform on a pandas column object directly impacts the pandas table itself.  I wonder if this is a bug in pandas - I'm going to open an issue.

PR #278, opened 2019-06-28
Upgrading numpy version

mortonjt:
This will allow us to upgrade the skbio version


mortonjt:
ugh .. we are getting a bunch of outdated errors from bokeh + pandas nonsense.

mortonjt:
ok, turns out that both matplotlib and bokeh haven been updated.

The plots look fine, but the order of which the objects have been plotted has changed.

image

image

mortonjt:
Tests should be passing now.  We will want to deprecate all visualizations in gneiss in the near future.  The heatmap dendrograms have been pushed to calour and the radial trees have been pushed to empress.

mortonjt:
Alright.  Merging.

Issue #287, opened 2021-04-09
Compatiability with xarray


mortonjt:
There is the scenario where we will need to compute the ILR transform across an entire tensor.  
Pull requests will be coming up shortly.

PR #288, opened 2021-04-09
WIP adding xarray match tips function

mortonjt:
This addresses issue #287 so that we can compute ILR transforms on tensors :)


mortonjt:
welp, it looks like travis is down; so we'll need to opt for another service like github actions ...

Issue #290, opened 2021-07-08
BUG: match appears to be broken with biom tables


mortonjt:
Not sure exactly what is going on here, but it looks like the the sample ids don't align with biom tables.

You can see this with the intermittent hypoxia dataset : https://qiita.ucsd.edu/study/description/10422
Below is some example to reproduce behavior (file paths do need to be changed though).
```python
t_md = pd.read_table('test_metadata.txt', index_col=0)
t_biom = biom.load_table('test.biom')
t_biom, t_md = match(t_biom, t_md)
print(t_biom.ids())
print(t_md.index)
```
```
['10422.16.F.3', '10422.17.F.12', '10422.28.F.1', '10422.31.F.9', ...
```
```
['10422.25.F.3', '10422.23.F.4', '10422.30.F.2', '10422.12.F.14', ...
```
You can see that the ids are *not* matched

PR #291, opened 2021-07-08
Fixing sparse matching

mortonjt:
This fixes the issue with sparse matching with #290 .  We still need to fix travis to get this in.


mortonjt:
I've also updated both matching methods so that the set ops don't add any additional randomness.
Turns out that set ops are inherently unstable, no random seeds can make the set ordering reproducible.
So after taking intersections, these sets are converted to lists and sorted.

